PR #10, opened 2021-04-10
Migrate to flutter 2

jonataslaw:
Hi, I noticed that the project became incompatible with one of our projects after the Flutter 2 update, so I migrated it

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -21,7 +21,7 @@ void main() {
   ///
   /// Aside from this special use case, any other context property that needs
   /// to be defined (or re-defined) can be done.
-  Posthog.setContext({
+  Posthog().setContext({
     'device': {
       'token': 'testing',
     }
@@ -33,7 +33,7 @@ void main() {
 class MyApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
-    Posthog.screen(
+    Posthog().screen(
       screenName: 'Example Screen',
     );
     return MaterialApp(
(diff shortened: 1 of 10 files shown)


lib/src/posthog.dart
```diff
@@ -30,9 +28,9 @@ class Posthog {
   }
 
   Future<void> capture({
-    @required String eventName,
-    Map<String, dynamic> properties,
-    Map<String, dynamic> options,
+    required String eventName,
+    required Map<String, dynamic> properties,
```
yakkomajuri, line 32: this seems to have diverged from lib/src/posthog_method_channel.dart where properties are correctly assigned as optional

example/lib/main.dart
```diff
@@ -61,49 +61,52 @@ class MyApp extends StatelessWidget {
...
             Spacer(),
             Center(
-              child: FlatButton(
+              child: TextButton(
                 child: Text('Enable'),
                 onPressed: () async {
-                  await Posthog.enable();
-                  Posthog.capture(eventName: 'Enabled capturing events!');
+                  await Posthog().enable();
+                  Posthog().capture(
```
yakkomajuri, line 98: Is this standard Dart indentation? (I don't know so I'm truly asking)

yakkomajuri (commented):
Hey, thank you so much for doing this! We don't have a dedicated Dart/Flutter dev hence it took a while to get around to this. 

Left a few minor comments but I'm still testing this out

yakkomajuri:
@jonataslaw this worked for me and all the changes make sense! Thanks again for doing this. A few minor comments otherwise happy to merge.

example/lib/main.dart
jonataslaw: Yeah, it's pretty odd for those who are used to other languages, but it's the default, I even used dart format in this file before saving the changes

lib/src/posthog.dart
jonataslaw: You're right, I'm going to make it nullable instead of required

jonataslaw:
I just changed properties to nullable to be consistent with posthog_method_channel now

yakkomajuri (approved):
Thanks again! Will get this in but I don't have access to publish a new version, will look to get access to this internally

fuziontech:
lib pushed to pub.dev! 
https://pub.dev/packages/posthog_flutter

Available version is now 2.0.0

PR #16, opened 2021-06-29
fix: flutter web identify getAnonymousId error.

shinyo17:
1. I found that Posthog().identify not works in Flutter web. The reason is calling distinctId. So I changed distinctId to userId.
2. I found that Posthog().getAnonymousId not works in Flutter web. It makes error like this.
```
Error: PlatformException(error, NoSuchMethodError: 'user'
method not found
Receiver: Instance of 'NativeJavaScriptObject'
Arguments: [], null, null)
```
In posthog-js, getting the anonymous Id is able to call the method "get_distinct_id". So, I changed the code to call method "get_distinct_id".

https://posthogusers.slack.com/archives/CT7HXDEG3/p1624946446377900

I sended at Slack about the error.

### DIFF
--- a/lib/src/posthog_web.dart
+++ b/lib/src/posthog_web.dart
@@ -3,7 +3,6 @@ import 'dart:js';
 import 'package:flutter/services.dart';
 import 'package:flutter_web_plugins/flutter_web_plugins.dart' show Registrar;
 
-
 class PosthogWeb {
   static void registerWith(Registrar registrar) {
     final MethodChannel channel = MethodChannel(
@@ -20,7 +19,7 @@ class PosthogWeb {
     switch (call.method) {
       case 'identify':
         analytics.callMethod('identify', [
-          call.arguments['distinctId'],
+          call.arguments['userId'],
           JsObject.jsify(call.arguments['properties']),
         ]);
         break;
@@ -42,8 +41,7 @@ class PosthogWeb {
         ]);
         break;
       case 'getAnonymousId':
-        final user = analytics.callMethod('user');
-        final anonymousId = user.callMethod('anonymousId');
+        final anonymousId = analytics.callMethod('get_distinct_id');
         return anonymousId;
       case 'reset':
         analytics.callMethod('reset');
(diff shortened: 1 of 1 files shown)


lib/src/posthog_web.dart
```diff
@@ -20,7 +19,7 @@ class PosthogWeb {
     switch (call.method) {
       case 'identify':
         analytics.callMethod('identify', [
-          call.arguments['distinctId'],
```
fuziontech, line 23: I think keeping it as `distinctId` keeps it aligned with our JS library
https://github.com/PostHog/posthog-js/blob/master/src/posthog-core.js#L810

Have you tried passing your `userId` as the `distinctId` parameter?

lib/src/posthog_web.dart
```diff
@@ -42,8 +41,7 @@ class PosthogWeb {
         ]);
         break;
       case 'getAnonymousId':
-        final user = analytics.callMethod('user');
-        final anonymousId = user.callMethod('anonymousId');
```
fuziontech, line 46: Nice! Great catch here :)

fuziontech (changes requested):
Thank you so much for the PR! Nice catch on the `get_distinct_id` call.

Can you try passing your `userId` as `distinctId` parameter in `identify` to see if that works?

lib/src/posthog_web.dart
shinyo17: Thank you for giving me a feedback!
I changed distinctId to userId because of the default setting of Posthog().identify.
To use  Posthog().identify, there is a required parameter "userId".
So I have to use like this. 
```
Posthog().identify(userId: "Shinyoung Tyga Lee");
```

shinyo17: @fuziontech 
Thank you for giving me a feedback!
I changed distinctId to userId because of the default setting of Posthog().identify.
To use  Posthog().identify, there is a required parameter "userId".
So I have to use like this.  I think that using distinctId should make change about default setting of required parameter "userId" name.
```
Posthog().identify(userId: "Shinyoung Tyga Lee");
```

fuziontech: You are totally right! https://github.com/PostHog/posthog-flutter/blob/7beff569182860749afcd802d9d8155c464b4ed0/lib/src/posthog.dart#L19
Your changes look great and make this more consistent.

fuziontech (approved):
Changes look good!

shinyo17:
Thank you!

fuziontech:
No prob! give `2.0.3` a try to make sure everything is working 👍

PR #19, opened 2021-11-11
Check that current screen is not null before setting it in the properties

savy-91:


fuziontech (approved):
Thank you for your contribution @savy-91 !

PR #25, opened 2022-11-03
feat: Add enable and disable to web implementation

fdaugs:
# Changes
This PR adds support for `enable` and `disable` for Flutter Web.

# Implementation
Following the existing implementation the calls will be delegated to the Javascript Client Library. `enable` maps to `opt_in_capturing` and disable to `opt_out_capturing`.

This Change does not belog to an open issue. But seemed like a pretty easy addition to me.


fuziontech (approved):
👍

Issue #33, opened 2023-03-25
ARC Semantic Issue (Xcode): No visible @interface for 'PosthogFlutterPlugin' declares the selector 'reloadFeatureFlags:result:'


coder-Aayush:
I have followed getting started guide and added configuration code for iOS. But facing issue while compiling.
Flutter version: 3.7.5 

posthog_flutter: 3.0.4

<details>
  <summary>Configuration</summary>

```plist

<key>com.posthog.posthog.API_KEY</key>
	<string>****</string>
	<key>com.posthog.posthog.POSTHOG_HOST</key>
	<string>https://app.posthog.com</string>
	<key>com.posthog.posthog.TRACK_APPLICATION_LIFECYCLE_EVENTS</key>
	<false/>

```
</details>

<details>
  <summary>Flutter Doctor</summary>

```dart
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.7.5, on macOS 13.1 22C65 darwin-arm64, locale en-NP)
[✓] Android toolchain - develop for Android devices (Android SDK version 32.1.0-rc1)
[✓] Xcode - develop for iOS and macOS (Xcode 14.2)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.2)
[✓] VS Code (version 1.76.1)
[✓] Connected device (3 available)
[✓] HTTP Host Availability

• No issues found!
```
</details>

<details>
  <summary>Error Log</summary>

```dart
Launching lib/main_development.dart on iPhone 14 in debug mode...
main_development.dart:1
Xcode build done.                                            6.0s
Failed to build iOS app
ARC Semantic Issue (Xcode): No visible @interface for 'PosthogFlutterPlugin' declares the selector 'reloadFeatureFlags:result:'
/Users/.pub-cache/hosted/pub.dev/posthog_flutter-3.0.4/ios/Classes/PosthogFlutterPlugin.m:119:10

Could not build the application for the simulator.
Error launching application on iPhone 14.
Exited
```
</details>

ajsharp:
Seeing same thing here.

ajsharp:
From posthog [community support](https://posthogusers.slack.com/archives/C01GLBKHKQT/p1679609870172029?thread_ts=1678805719.691969&cid=C01GLBKHKQT):

> hi there! please roll back to pre 3.0.0 since that includes the breaking changes for feature flags.

dio-ralali:
[...]

Issue #31, opened 2023-03-14
iOS build fails with ARC Semantic Issue (Xcode) on Version 3.0.4


Wizzel1:
I have just updated posthog for my iOS App but when building the App  I am getting an error:

```
Failed to build iOS app
Could not build the precompiled application for the device.
ARC Semantic Issue (Xcode): No visible @interface for 'PosthogFlutterPlugin' declares the selector 'reloadFeatureFlags:result:'
```

dio-ralali:
have you found any solutions? I facing the same issue on my project.

<img width="1072" alt="CleanShot 2023-04-05 at 14 28 55@2x" src="https://user-images.githubusercontent.com/86591816/230011588-f9fd9995-e267-4b44-bbfd-ba2c3e9e624a.png">

Wizzel1:
@dio-ralali No I haven't, I have downgraded the package back to 2.0.3 for now

dio-ralali:
@Wizzel1 thanks. downgrading the package works for me too.

PR #35, opened 2023-06-01
fix: iOS build issue

benjackwhite:
The recently added Feature flags support had a bug.

Closes https://github.com/PostHog/posthog-flutter/issues/33
Closes https://github.com/PostHog/posthog-flutter/issues/31

## Changs
* Fixes the bug
* Adds a CI workflow for building the example repo to at least check that everything is correct


neilkakkar (approved):
nice, thanks! looks like code 👀

PR #43, opened 2023-07-18
Fix version constraint in `pubspec.yaml`

bartekpacia:
Dart version is set to 2.12, which shipped in Flutter 2.0 ([src](https://developers.googleblog.com/2021/03/announcing-flutter-2.html)).

It's a good practice to keep these two in sync.


marandaneto:
Closing in favor of https://github.com/PostHog/posthog-flutter/pull/52/files
It's okay to keep a lower API compatibility as long as its working for older and newer versions.

PR #48, opened 2023-09-13
Update README.md

Kiruel:
Update the POSTHOG_HOST to reflect the new posthog cloud.


Kiruel:
Or maybe do like the official documentation:
```<ph_instance_address>```
And add a list of address somewhere.

Kiruel:
I will close this issue, but maybe we should add a list of addresses. And where we can get it.

PR #52, opened 2023-10-09
Update Flutter to be compatible with Dart v3

marandaneto:
Closes https://github.com/PostHog/posthog-flutter/pull/43
Closes https://github.com/PostHog/posthog-flutter/pull/42
Dart 3 is out since May https://dart.dev/guides/whats-new#may-10-2023-30-release
Flutter uses Dart 3 already
Right now people cannot use our SDK since there are dependency boundaries.


benjackwhite (approved):
👍

PR #54, opened 2023-10-10
Update tooling to latest version

marandaneto:
Reason: 

* Otherwise you can't use the latest Android tooling or latest supported Java (who still uses Java 7 only?)
* Deprecations

TODO:
* Migrate Flutter engine plugin to min v2 or v3 and increase the Flutter min version (not on this PR).


android/src/main/java/com/posthog/posthog_flutter/PosthogFlutterPlugin.java
```diff
@@ -38,6 +35,7 @@ public class PosthogFlutterPlugin implements MethodCallHandler, FlutterPlugin {
   static HashMap<String, Object> appendToContextMiddleware;
 
   /** Plugin registration. */
+  @SuppressWarnings("deprecation")
```
marandaneto, line 38: Migrate to the new plugin

benjackwhite (approved):
All looks fine to me

PR #56, opened 2023-11-01
Fix an issue in the example

ueman:
Calls like `Posthog().screen(screenName: 'Example Screen');` should not be done in a widgets build method since the build method is potentially called multiple times. Therefore, this could lead to wrongly recorded analytics with too many screen views. Additionally, the `PosthogObserver` is already recording that exact screen view.

In this specific example, it might actually be acceptable, though, since it's the root widget. However, an example shouldn't be this nuanced, so it's better to remove it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,11 @@ void main() => runApp(MyApp());
 class MyApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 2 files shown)


marandaneto (approved):
I agree, thanks.

PR #72, opened 2024-01-05
ref: clean old files

marandaneto:

Release log:

> Total compressed archive size: 85 KB.
> Validating package...
> Package validation found the following potential issue:
> * 2 checked-in files are ignored by a `.gitignore`.
>   Previous versions of Pub would include those in the published package.
>   
>   Consider adjusting your `.gitignore` files to not ignore those files, and if you do not wish to
>   publish these files use `.pubignore`. See also dart.dev/go/pubignore
>   
>   Files that are checked in while gitignored:
>   
>   example/android/app/src/debug/AndroidManifest.xml
>   example/ios/Podfile.lock

They don't need to be checked in.
Strange enough, `example/android/app/src/debug/AndroidManifest.xml` isn't ignored and has to be checked in, likely a bug in the analyzer.

_#skip-changelog_

## :pencil: Checklist


example/.gitignore
```diff
@@ -38,10 +38,6 @@ app.*.symbols
 # Obfuscation related
 app.*.map.json
 
-# Android Studio will place build artifacts here
```
marandaneto, line 41: Its always `build/` and its already ignored, likely an old gitignore file

example/ios/Podfile.lock
```diff
@@ -1,29 +0,0 @@
-PODS:
```
marandaneto, line 1: this is auto generated anyways

Issue #85, opened 2024-02-29
Missing Events Immediately After Startup


jakobhec:
### Version

4.0.1

### Steps to Reproduce

Capture an event very early after the app starts up. For example, here we do it right after `WidgetsFlutterBinding.ensureInitialized()`.

```dart
Future<void> main() async {
  WidgetsBinding widgetsBinding = WidgetsFlutterBinding.ensureInitialized();
  Posthog().capture(eventName: 'dummy_event');
  ...
```

### Expected Result

I would expect all events to show up on my dashboard, regardless of when I capture them in the app's lifecycle.

### Actual Result

Events within the first second after the app starts up are not always captured. A rough estimate would be that 90% of the `dummy_event` in the example above are missed, only 10% appear on my dashboard.

I would expect 100% of the events to appear on my dashboard. 

This works for subsequent events (e.g., when I log a button tap on the second screen).

marandaneto:
@jakobhec can you enable the `debug` option and check the logs? I'd be interested to know if there's anything strange.

marandaneto:
@jakobhec your test is Android or iOS?

marandaneto:
@jakobhec did you check it?

marandaneto:
@jakobhec can you reproduce this issue? can you provide a [MRE](https://en.wikipedia.org/wiki/Minimal_reproducible_example)? I cannot reproduce myself sadly, everything works as expected.

marandaneto:
> @marandaneto If the app is opened, 15 events are triggered by a user, and then the app is closed, will PostHog send those events? Currently thinking about the MRE

Yes, by default the SDK will only send after it's accumulated at least 20 events or the app is idle for at least 30 seconds.

marandaneto:
> @marandaneto just to clarify: in the example there were less than 20 events and the app was not idle for at least 30 secondes. Shouldn't that mean that nothing will be sent?
> 
> Or does posthog also trigger on app close?

It's either at least 20 events or after 30 seconds the app is opened, not idle, sorry wrong wording.

marandaneto:
> @marandaneto thanks for clarifying.
> 
> So if a user opens an app for the first time, sees the onboarding flow, slides through the first 3 onboarding screens (triggering < 20 events in total) and then leaves the app after 15 seconds. Will Posthog send events?

yes, after 30 seconds if the app is open (it can be in the background), but not closed (forcefully killed from recent apps), events should have been sent.

[...]

PR #92, opened 2024-03-07
chore: add flush method

marandaneto:
Closes https://github.com/PostHog/posthog-flutter/issues/85#issuecomment-1983475804

## :pencil: Checklist

### DIFF
--- a/lib/src/posthog_flutter_web_handler.dart
+++ b/lib/src/posthog_flutter_web_handler.dart
@@ -90,6 +90,10 @@ Future<dynamic> handleWebMethodCall(MethodCall call, JsObject context) async {
         call.arguments['key'],
       ]);
       break;
+    case 'flush':
+      // not supported on Web
+      // analytics.callMethod('flush');
+      break;
     default:
       throw PlatformException(
         code: 'Unimplemented',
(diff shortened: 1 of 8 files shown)


lib/src/posthog_flutter_web_handler.dart
```diff
@@ -90,6 +90,10 @@ Future<dynamic> handleWebMethodCall(MethodCall call, JsObject context) async {
         call.arguments['key'],
       ]);
       break;
+    case 'flush':
+      // not supported on Web
+      // analytics.callMethod('flush');
```
marandaneto, line 95: @pauldambra is there an alternative to flush on the web?

daibhin: Looks like there is no equivalent: https://github.com/PostHog/posthog-js/issues/205#issuecomment-1022028337

daibhin (approved):
Looks good other than the open question on web

lib/src/posthog_flutter_web_handler.dart
pauldambra: No, I'd guess those comments are up to date

marandaneto: yeah didn't find in the code either, will merge as it is.

PR #95, opened 2024-04-02
chore: Allow overriding the route filtering using a ctor param `routeFilter`

marandaneto:
https://twitter.com/luke_pighetti/status/1774906968886366545

Using a fake `PosthogFlutterPlatformInterface`, it was the easiest way to test it and not break or add any public APIs

## :pencil: Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,9 @@ jobs:
         working-directory: ./example
         run: flutter pub get
 
+      - name: Test
+        run: flutter test
+
       - name: Build iOS
         working-directory: ./example
         run: flutter build ios --simulator --no-codesign
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 ## Next
 
-- Allow overriding the route filtering usint a ctor param `routeFilter` ([#95](https://github.com/PostHog/posthog-flutter/pull/95))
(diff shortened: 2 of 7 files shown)


marandaneto:
cc @lukepighetti

daibhin (approved):

lukepighetti:
one way to make this non-breaking for existing customers is to make isTrackableRoute a class constructor argument with a default that can be overridden

you could provide the current one as a default and the new one as an alternative exported from the package

marandaneto:
> one way to make this non-breaking for existing customers is to make isTrackableRoute a class constructor argument with a default that can be overridden
> 
> you can see that pattern here in `firebase_analytics` [firebase/flutterfire@`master`/packages/firebase_analytics/firebase_analytics/lib/observer.dart#L17-L22](https://github.com/firebase/flutterfire/blob/master/packages/firebase_analytics/firebase_analytics/lib/observer.dart?rgh-link-date=2024-04-02T12%3A04%3A25Z#L17-L22)
> 
> i bet if you copy paste this but rename the typedef to `PosthogRouteFilter` folks will be able to use the same filter on firebase_analytics and posthog
> 
> you could provide the current one as a default and the new one as an alternative exported from the package
> 
> related discussion: [twitter.com/ue_man/status/1775125806789894437](https://twitter.com/ue_man/status/1775125806789894437)

Makes sense, this [tweet](https://x.com/ue_man/status/1775004768781439320?s=20) convinced me, better to allow the customization, fixes are pushed, thanks for the review.

PR #126, opened 2024-12-04
chore: do not send repeated snapshots

marandaneto:
do not send repeated snapshots if nothing changed

running sample and being sure that the screen should or not match the last snapshot

## :pencil: Checklist

### DIFF
--- a/lib/src/posthog_flutter_web_handler.dart
+++ b/lib/src/posthog_flutter_web_handler.dart
@@ -102,6 +102,18 @@ Future<dynamic> handleWebMethodCall(MethodCall call, JsObject context) async {
       // not supported on Web
       // analytics.callMethod('close');
       break;
+    case 'sendMetaEvent':
+      // not supported on Web
+      // Flutter Web uses the JS SDK for Session replay
+      break;
+    case 'sendFullSnapshot':
+      // not supported on Web
+      // Flutter Web uses the JS SDK for Session replay
+      break;
+    case 'isSessionReplayActive':
+      // not supported on Web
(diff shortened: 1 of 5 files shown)


lib/src/posthog_widget_widget.dart
```diff
@@ -75,17 +77,32 @@ class PostHogWidgetState extends State<PostHogWidget> {
...
 
     Uint8List pngBytes = byteData.buffer.asUint8List();
     imageInfo.image.dispose();
 
+    if (pngBytes.isEmpty) {
+      printIfDebug('Error: Failed to convert image byte data to Uint8List.');
+      return;
+    }
+
+    if (const PHListEquality().equals(pngBytes, _lastSnapshot)) {
```
marandaneto, line 99: hard to tell the performance impact of that, can move to a background thread on the native side if needed

lib/src/posthog_flutter_web_handler.dart
```diff
@@ -102,6 +102,18 @@ Future<dynamic> handleWebMethodCall(MethodCall call, JsObject context) async {
...
+      // Flutter Web uses the JS SDK for Session replay
+      break;
+    case 'sendFullSnapshot':
+      // not supported on Web
+      // Flutter Web uses the JS SDK for Session replay
+      break;
+    case 'isSessionReplayActive':
+      // not supported on Web
+      // Flutter Web uses the JS SDK for Session replay
+      break;
```
marandaneto, line 116: flyby fix, missed it here

Issue #140, opened 2024-12-22
bug masks out of sync when moving too fast.


thisames:
### Version

4.9.0

### Steps to Reproduce

I did any tests and i notice that the bigger the throttleDelay, more the masking frame is losed

![image](https://github.com/user-attachments/assets/dc360a91-7ad8-4960-839d-c905d443cd8b)

### Expected Result

masking working normally

### Actual Result

masking not working normally

thisames:
I already investigate a possible solution for this.

marandaneto:
> I already investigate a possible solution for this.

I could not even reproduce this, mind opening a draft PR then?

thisames:
> > I already investigate a possible solution for this.
> 
> I could not even reproduce this, mind opening a draft PR then?

sorry, after investigating further, it wasn't the throttleDelay that was influencing the masked frames, but rather the way we called the async functions in the screenshot process.

If you want to reproduce this, just move the screen quickly, scroll quickly, etc.

I've already fixed this and fixed it in depth, there are just a few details left.

I'm going to open a draft-pr

PR #147, opened 2025-01-08
feat: Refactor screenshot capture for improved performance and reliability

thisames:
This commit refactors the `captureScreenshot` method in `ScreenshotCapturer`
to improve performance and reliability. The changes include:

-   **Synchronous Setup:** The method now performs synchronous operations first
    (getting context, render object, calculating pixel ratio, and getting the
    initial image) before executing the main process asynchronously. This
    ensures that the initial setup is done on the main UI thread, as required.
    
-   **Asynchronous Processing:** The main image processing and masking logic
    are now executed asynchronously using a `Future` and `Completer`.
    
    
#140 

## :pencil: Checklist

### DIFF
--- a/lib/src/posthog_widget_widget.dart
+++ b/lib/src/posthog_widget_widget.dart
@@ -3,6 +3,7 @@ import 'dart:async';
 import 'package:flutter/material.dart';
 import 'package:posthog_flutter/posthog_flutter.dart';
 import 'package:posthog_flutter/src/replay/mask/posthog_mask_controller.dart';
+import 'package:posthog_flutter/src/util/logging.dart';
 
 import 'replay/change_detector.dart';
 import 'replay/native_communicator.dart';
@@ -65,14 +66,11 @@ class PostHogWidgetState extends State<PostHogWidget> {
   }
 
   Future<void> _generateSnapshot() async {
-    final isSessionReplayActive =
-        await _nativeCommunicator?.isSessionReplayActive() ?? false;
-    if (!isSessionReplayActive) {
-      return;
-    }
-
+    // Ensure no asynchronous calls occur before this function,
+    // as it relies on a consistent state.
     final imageInfo = await _screenshotCapturer?.captureScreenshot();
     if (imageInfo == null) {
+      printIfDebug('Error: Failed to capture screenshot.');
       return;
     }
 
--- a/lib/src/replay/element_parsers/element_data.dart
+++ b/lib/src/replay/element_parsers/element_data.dart
@@ -16,24 +16,20 @@ class ElementData {
     children?.add(elementData);
   }
 
-  List<Rect> extractRects([bool isRoot = true]) {
-    List<Rect> rects = [];
-
-    if (!isRoot) {
-      rects.add(rect);
-    }
+  List<ElementData> extractRects([bool isRoot = true]) {
+    List<ElementData> rects = [];
 
     if (children != null) {
       for (var child in children ?? []) {
         if (child.children == null) {
-          rects.add(child.rect);
+          rects.add(child);
           continue;
         } else if ((child.children?.length ?? 0) > 1) {
           for (var grandChild in child.children ?? []) {
-            rects.add(grandChild.rect);
+            rects.add(grandChild);
           }
         } else {
-          rects.add(child.rect);
+          rects.add(child);
         }
       }
     }
--- a/lib/src/replay/mask/posthog_mask_controller.dart
+++ b/lib/src/replay/mask/posthog_mask_controller.dart
@@ -1,5 +1,3 @@
-import 'dart:async';
-
 import 'package:flutter/material.dart';
 import 'package:posthog_flutter/posthog_flutter.dart';
 import 'package:posthog_flutter/src/replay/element_parsers/element_data.dart';
@@ -10,6 +8,7 @@ import 'package:posthog_flutter/src/replay/element_parsers/element_data_factory.
 import 'package:posthog_flutter/src/replay/element_parsers/element_object_parser.dart';
 import 'package:posthog_flutter/src/replay/element_parsers/root_element_provider.dart';
 import 'package:posthog_flutter/src/replay/mask/widget_elements_decipher.dart';
+import 'package:posthog_flutter/src/util/logging.dart';
 
 class PostHogMaskController {
   late final Map<String, ElementParser> parsers;
@@ -32,15 +31,41 @@ class PostHogMaskController {
       PostHogMaskController._privateConstructor(
           Posthog().config?.sessionReplayConfig);
 
-  Future<List<Rect>?> getCurrentScreenRects() async {
+  /// Extracts a flattened list of [ElementData] objects representing the
+  /// renderable elements in the widget tree.
+  ///
+  /// This method traverses the tree of [ElementData] objects and returns a
+  /// list of elements that have no children or only one child.
+  ///
+  /// The method is designed to extract the elements that are directly
+  /// renderable on the screen.
+  ///
+  /// **Returns:**
+  ///   - `List<ElementData>`: A list of [ElementData] objects representing the
+  ///     renderable elements.
+  ///
+  List<ElementData>? getCurrentWidgetsElements() {
     final BuildContext? context = containerKey.currentContext;
 
     if (context == null) {
+      printIfDebug('Error: containerKey.currentContext is null.');
       return null;
     }
-    final ElementData? widgetElementsTree =
-        _widgetScraper.parseRenderTree(context);
 
-    return widgetElementsTree?.extractRects();
+    try {
+      final ElementData? widgetElementsTree =
+          _widgetScraper.parseRenderTree(context);
+
+      if (widgetElementsTree == null) {
+        printIfDebug('Error: widgetElementsTree is null after parsing.');
+        return null;
+      }
+
+      return widgetElementsTree.extractRects();
+    } catch (e) {
+      printIfDebug(
+          'Error during render tree parsing or rectangle extraction: $e');
+      return null;
+    }
   }
 }
--- a/lib/src/replay/screenshot/screenshot_capturer.dart
+++ b/lib/src/replay/screenshot/screenshot_capturer.dart
@@ -1,10 +1,14 @@
+import 'dart:async';
 import 'dart:math';
 import 'dart:typed_data';
 import 'dart:ui' as ui;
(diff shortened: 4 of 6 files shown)


lib/src/replay/screenshot/screenshot_capturer.dart
```diff
@@ -30,8 +34,8 @@ class ViewTreeSnapshotStatus {
 class ScreenshotCapturer {
```
thisames, line 34: what do you think about changing the class name? `ScreenshotCapturer` looks weird to me

lib/src/posthog_widget_widget.dart
```diff
@@ -65,14 +66,11 @@ class PostHogWidgetState extends State<PostHogWidget> {
   }
 
   Future<void> _generateSnapshot() async {
-    final isSessionReplayActive =
-        await _nativeCommunicator?.isSessionReplayActive() ?? false;
-    if (!isSessionReplayActive) {
-      return;
-    }
-
+    // Ensure no asynchronous calls occur before this function,
```
marandaneto, line 69: calling `isSessionReplayActive` is important even if its inside `captureScreenshot` because we don't want to pay the performance price of taking the screenshot if session replay is inactive

lib/src/replay/element_parsers/element_data.dart
```diff
@@ -16,24 +16,20 @@ class ElementData {
     children?.add(elementData);
   }
 
-  List<Rect> extractRects([bool isRoot = true]) {
-    List<Rect> rects = [];
-
-    if (!isRoot) {
-      rects.add(rect);
-    }
+  List<ElementData> extractRects([bool isRoot = true]) {
```
marandaneto, line 19:
```suggestion
  List<ElementData> extractRects({bool isRoot = true}) {
```

I think optionals are better than positionals

lib/src/replay/mask/posthog_mask_controller.dart
```diff
@@ -32,15 +31,41 @@ class PostHogMaskController {
...
+  /// list of elements that have no children or only one child.
+  ///
+  /// The method is designed to extract the elements that are directly
+  /// renderable on the screen.
+  ///
+  /// **Returns:**
+  ///   - `List<ElementData>`: A list of [ElementData] objects representing the
+  ///     renderable elements.
+  ///
+  List<ElementData>? getCurrentWidgetsElements() {
```
marandaneto, line 47:
```suggestion
  List<ElementData>? _getCurrentWidgetsElements() {
```

can be private I assume

lib/src/replay/mask/posthog_mask_controller.dart
```diff
@@ -32,15 +31,41 @@ class PostHogMaskController {
...
+      printIfDebug('Error: containerKey.currentContext is null.');
       return null;
     }
-    final ElementData? widgetElementsTree =
-        _widgetScraper.parseRenderTree(context);
 
-    return widgetElementsTree?.extractRects();
+    try {
+      final ElementData? widgetElementsTree =
+          _widgetScraper.parseRenderTree(context);
```
marandaneto, line 57:
```suggestion
      final widgetElementsTree =
          _widgetScraper.parseRenderTree(context);
```

the type is inferred already

lib/src/replay/screenshot/screenshot_capturer.dart
marandaneto: find to rename it, its only used internally

lib/src/replay/screenshot/screenshot_capturer.dart
```diff
@@ -47,37 +51,33 @@ class ScreenshotCapturer {
...
-
   Future<Uint8List?> _getImageBytes(ui.Image img) async {
-    final ByteData? byteData =
-        await img.toByteData(format: ui.ImageByteFormat.png);
-    if (byteData == null || byteData.lengthInBytes == 0) {
-      printIfDebug('Error: Failed to convert image to byte data.');
+    try {
+      final ByteData? byteData =
+          await img.toByteData(format: ui.ImageByteFormat.png);
+      if (byteData == null) {
```
marandaneto, line 58: why removing the `byteData.lengthInBytes == 0` check?

lib/src/replay/screenshot/screenshot_capturer.dart
```diff
@@ -47,37 +51,33 @@ class ScreenshotCapturer {
...
+      return byteData.buffer.asUint8List();
+    } catch (e) {
+      printIfDebug('Error converting image to byte data: $e');
       return null;
     }
-    return byteData.buffer.asUint8List();
   }
 
-  Future<ImageInfo?> captureScreenshot() async {
+  Future<ImageInfo?> captureScreenshot() {
```
marandaneto, line 69: the function has to be async if it returns a Future, right?

lib/src/replay/screenshot/screenshot_capturer.dart
thisames: sorry, I had copy the function of old version mine

lib/src/posthog_widget_widget.dart
thisames: yes, I put this in 

https://github.com/PostHog/posthog-flutter/blob/25a3690d6fe4cb5d53516c7b70161fffaabd106e/lib/src/replay/screenshot/screenshot_capturer.dart#L111-L117

what do you think?

lib/src/replay/screenshot/screenshot_capturer.dart
```diff
@@ -86,79 +86,135 @@ class ScreenshotCapturer {
...
+        final isSessionReplayActive =
+            await _nativeCommunicator.isSessionReplayActive();
+        if (!isSessionReplayActive) {
+          return;
+        }
 
-      statusView.imageBytes = pngBytes;
+        final recorder = ui.PictureRecorder();
+        final canvas = Canvas(recorder);
+        final image = await syncImage;
```
marandaneto, line 120:
```suggestion
        final image = await renderObject.toImage(pixelRatio: pixelRatio);
```

would that work?

lib/src/replay/screenshot/screenshot_capturer.dart
```diff
@@ -86,79 +86,135 @@ class ScreenshotCapturer {
...
+    final Completer<ImageInfo?> completer = Completer<ImageInfo?>();
+
     try {
       final srcWidth = renderObject.size.width;
       final srcHeight = renderObject.size.height;
       final pixelRatio =
           _getPixelRatio(srcWidth: srcWidth, srcHeight: srcHeight);
 
-      final ui.Image image = await renderObject.toImage(pixelRatio: pixelRatio);
+      final syncImage = renderObject.toImage(pixelRatio: pixelRatio);
```
marandaneto, line 97: well this is misleading because its not sync, since returns a future and you don't `await`.

marandaneto:
nice, almost there.

lib/src/replay/screenshot/screenshot_capturer.dart
thisames: Not necessarily, a function does not need to be explicitly marked as async to return a Future.

In our case, we need to work synchronously with the function. Then we execute the main process asynchronously as a "task". Managing this with the Completer. When this is ready, we retrieve the Completer and return the Future.

This way, the frame image and the masks always move in the main thread.

marandaneto: true that, nice

lib/src/replay/screenshot/screenshot_capturer.dart
thisames: with 

```dart
final image = await renderObject.toImage(pixelRatio: pixelRatio);
```

the renderObject reference may have changed in the main thread, so it would be out of sync

marandaneto: can we do the whole process inside of the Future then? so its all done in sequence

thisames: But if you have a solution to the whole process in the Future block, I would like to see it, maybe I missed something or something is beyond my ability haha

marandaneto: everything right now is being done on the main thread, Dart is single-threaded, unless you use [background workers](https://dart.dev/language/concurrency#background-workers) which is not available in all Dart/Flutter versions.
Maybe you are misunderstanding the concept of async/await with multithreading?

marandaneto: @thisames something like this [`9992590` (#150)](https://github.com/PostHog/posthog-flutter/pull/150/commits/99925900e4d6d8eb918473b8dba6484ca96eff81)
can you check if this works?

thisames: sure, I tested with this code, see the result

https://github.com/user-attachments/assets/7f6aae7a-7dbd-4df3-afdb-e2746635d647

thisames: about isolate, I think it wouldn't work because `RenderRepaintBoundary` `toImage()` makes calls that depend on the Flutter engine to draw on a GPU surface (or internal buffer). This process all occurs within the rendering engine, which is already linked to the main isolate. I believe that calling a secondary isolate for this wouldn't work...

lib/src/replay/mask/posthog_mask_controller.dart
thisames: we call it here

https://github.com/PostHog/posthog-flutter/blob/25a3690d6fe4cb5d53516c7b70161fffaabd106e/lib/src/replay/screenshot/screenshot_capturer.dart#L103-L106

lib/src/replay/screenshot/screenshot_capturer.dart
marandaneto: yep, i was just explaining the usage of background workers and isolates, but this has to be done on the main isolate.

marandaneto:
@thisames if you apply the suggested changes, we can get this merged and released :)

thisames:
I uploaded the small changes to the code review

What do you think about leaving the captureScreenshot method like this for now? It's a lesson to try to solve it with isolate or another better way. I think these details are a bit complex and the sooner we unlock these buggy masks the better.

marandaneto:
@thisames ok, can you add a changelog entry and make it ready for review?

marandaneto (approved):

PR #158, opened 2025-02-05
fix: isSessionReplayActive returns false by default for flutter web

marandaneto:

> Error sending full snapshot to native: TypeError: null: type 'Null' is not a subtype of type 'FutureOr<bool>'

Potential fix for https://posthoghelp.zendesk.com/agent/tickets/24603 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/26602)

## :pencil: Checklist

### DIFF
--- a/lib/src/posthog_flutter_web_handler.dart
+++ b/lib/src/posthog_flutter_web_handler.dart
@@ -121,7 +121,7 @@ Future<dynamic> handleWebMethodCall(MethodCall call, JsObject context) async {
     case 'isSessionReplayActive':
       // not supported on Web
       // Flutter Web uses the JS SDK for Session replay
-      break;
+      return false;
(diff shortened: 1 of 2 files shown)


lib/src/posthog_flutter_web_handler.dart
```diff
@@ -121,7 +121,7 @@ Future<dynamic> handleWebMethodCall(MethodCall call, JsObject context) async {
     case 'isSessionReplayActive':
       // not supported on Web
       // Flutter Web uses the JS SDK for Session replay
-      break;
+      return false;
```
marandaneto, line 124: since this is handled by the JS SDK directly, we just return false here.

pauldambra (approved):
💖

Issue #51, opened 2023-10-09
Desktop support


marandaneto:
https://posthog.com/questions/flutter-for-windows-support

The Flutter SDK depends on the Android and iOS SDK.
Ideally, the SDK would also have a pure Dart client that works on Desktop.

marandaneto:
A workaround, for now, would be to create a Flutter bridge and call a Native SDK via FFI, eg, [Rust SDK](https://github.com/PostHog/posthog-rs)

CaptainDario:
Is there anything planned in regard to this?
I would like to switch to this plugin but I need desktop support.

marandaneto:
@CaptainDario unfortunately not yet, but please upvote for gathering interest.

We'd need to write a pure Dart SDK (not depending on the Native SDKs such as Android and iOS for Desktop) or embed a Native SDK for Desktop, such as the Rust SDK.

However, adding macOS support should be easier since the iOS SDK supports a macOS target.

Which targets do you need? Win/Lin/macOS or all of them?

CaptainDario:
Thank you for the quick response! I upvoted this issue.

I would need support for all platforms. But I also see quite an opportunity in supporting these platforms for this package. Because firebase does not support desktop, and everybody needing desktop support will look somewhere else (just like me)

marandaneto:
A quick workaround, for now, is to just call the [API directly](https://posthog.com/tutorials/api-capture-events) until we provide first-class support for Flutter Desktop.

CaptainDario:
I totally missed the direct API access, this seems like a good solution.

CaptainDario:
@marandaneto How can I overwrite the default timestamp?
When I try to set it the events are not registered anymore in the web UI.

This works and the event is registered in the UI
``` dart
Map response = jsonDecode((await http.post(
    Uri.parse('https://eu.posthog.com/capture/'),
    headers: {
      "Content-Type": "application/json",
    },
    body: jsonEncode({
      "api_key": Env.POSTHOG_API_KEY,
      "event": eventName,
      "properties" : completeProps,
      //"timestamp" : DateTime(2023, 10, 5).toIso8601String(),
    })
  )).body);
```

this even is not registered.
``` dart
Map response = jsonDecode((await http.post(
    Uri.parse('https://eu.posthog.com/capture/'),
    headers: {
      "Content-Type": "application/json",
    },
    body: jsonEncode({
      "api_key": Env.POSTHOG_API_KEY,
[...]

PR #173, opened 2025-04-07
chore: windows and linux noop

marandaneto:
https://github.com/PostHog/posthog-flutter/issues/51#issuecomment-2770630838

## :pencil: Checklist

### DIFF
--- a/lib/src/posthog_flutter_io.dart
+++ b/lib/src/posthog_flutter_io.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:flutter/services.dart';
 import 'package:posthog_flutter/src/util/logging.dart';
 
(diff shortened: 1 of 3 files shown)


lib/src/posthog_flutter_io.dart
```diff
@@ -9,8 +11,16 @@ class PosthogFlutterIO extends PosthogFlutterPlatformInterface {
   /// The method channel used to interact with the native platform.
   final _methodChannel = const MethodChannel('posthog_flutter');
 
+  bool _isSupportedPlatform() {
+    return !(Platform.isLinux || Platform.isWindows);
+  }
```
rafaeelaudibert, line 16: In the same way `isLinux` is a getter, maybe this should be a getter too?

rafaeelaudibert (approved):
I'm not positive this should be a patch version. Maybe we should make it a minor so that people are 100% aware that we don't support native windows/linux?

Both this and my comment are just questions, feel free to merge as is

lib/src/posthog_flutter_io.dart
marandaneto: This should be a method since it checks different getters and returns a value based on conditions.
A getter is just an encapsulation thing that accesses a variable.

marandaneto:
> I'm not positive this should be a patch version. Maybe we should make it a minor so that people are 100% aware that we don't support native windows/linux?
> 
> Both this and my comment are just questions, feel free to merge as is

We don't support windows and linux, otherwise they would be listed [here](https://github.com/PostHog/posthog-flutter/blob/7226e0114ebd9b18ce944fdd9980ccc602080726/pubspec.yaml#L43)
This is more of an issue when you are running tests on a linux/windows machine

PR #174, opened 2025-04-09
fix: check if image size is valid before sending snapshot

marandaneto:
https://x.com/filippkowalski/status/1909872437006659795

## :pencil: Checklist


pauldambra (approved):
🙌

marandaneto:
> LG! Just out of curiosity, reported crash was already in a try-catch block though?

yes, I suspect its this line https://github.com/flutter/flutter/blob/2249c4dbbfa479fc6a4fe351515c8afab2faf606/engine/src/flutter/lib/ui/compositing.dart#L65

PR #179, opened 2025-04-28
Fixed wasm build and improved type safety

pedrovmc:
This change was necessary to modernize the web implementation of the PostHog Flutter plugin, replacing the deprecated `dart:js` API with the new `dart:js_interop` API. This fixes WASM compilation issues for Flutter applications using this plugin and improves type safety, reducing the possibility of runtime errors.

The change is crucial for:
1. Ensuring compatibility with newer versions of Flutter and Dart
2. Allowing the plugin to work correctly in WASM environments
3. Improving stability and security with stricter type checking

- Tested the plugin compilation for WASM using Flutter 3.19+
- Verified all core plugin functionalities in web environment:
  - Event capturing
  - User identification
  - Feature flags management
  - Reset and session management
- Compared the behavior of the new implementation with the previous implementation to ensure compatibility

## :pencil: Checklist
- [] I added tests to verify the changes.
- [] I updated the docs if needed.

### DIFF
--- a/lib/src/posthog_flutter_web_handler.dart
+++ b/lib/src/posthog_flutter_web_handler.dart
@@ -29,7 +29,7 @@ extension PostHogExtension on PostHog {
 
 // Acessar o posthog a partir do window
 @JS('window.posthog')
-external PostHog get posthog;
+external PostHog? get posthog;
 
 // Funções de conversão
 JSAny stringToJSAny(String value) {
@@ -41,7 +41,7 @@ JSAny boolToJSAny(bool value) {
 }
 
 JSAny mapToJSAny(Map<dynamic, dynamic> map) {
-  return map.jsify()!;
+  return map.jsify() ?? JSObject();
 }
 
 // Função para converter mapas de forma segura
@@ -71,7 +71,7 @@ Future<dynamic> handleWebMethodCall(MethodCall call) async {
       final userPropertiesSetOnce =
           safeMapConversion(args['userPropertiesSetOnce']);
 
-      posthog.identify(
+      posthog?.identify(
         stringToJSAny(userId),
         mapToJSAny(userProperties),
         mapToJSAny(userPropertiesSetOnce),
@@ -81,7 +81,7 @@ Future<dynamic> handleWebMethodCall(MethodCall call) async {
       final eventName = args['eventName'] as String;
       final properties = safeMapConversion(args['properties']);
 
-      posthog.capture(
+      posthog?.capture(
         stringToJSAny(eventName),
         mapToJSAny(properties),
       );
@@ -91,7 +91,7 @@ Future<dynamic> handleWebMethodCall(MethodCall call) async {
       final properties = safeMapConversion(args['properties']);
       properties['\$screen_name'] = screenName;
 
-      posthog.capture(
+      posthog?.capture(
         stringToJSAny('\$screen'),
         mapToJSAny(properties),
       );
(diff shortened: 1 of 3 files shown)


marandaneto:
awesome, thanks @pedrovmc 
i will review it tomorrow first thing.

pubspec.yaml
```diff
@@ -18,6 +18,7 @@ dependencies:
   plugin_platform_interface: ^2.0.2
   # plugin_platform_interface depends on meta anyway
   meta: ^1.3.0
+  web: ^1.1.0
```
marandaneto, line 21: this version requires Dart 3.4
so we need to bump this https://github.com/PostHog/posthog-flutter/blob/e44de573ea0aa2068d7c33a839d2c12130b1887c/pubspec.yaml#L10-L11
and then release the SDK as a new major version

marandaneto: min flutter 3.22 https://medium.com/flutter/io24-5e211f708a37 which is not super old.
can we use an older version of `web` so we can keep a better compatibility?
like https://pub.dev/packages/web/versions/0.4.2 uses Dart 3.2, Flutter 3.16

marandaneto: actually i dont see any import from `package:web`, do we actually need it?

lib/src/posthog_flutter_web_handler.dart
```diff
@@ -1,107 +1,180 @@
-// ignore: deprecated_member_use
-import 'dart:js';
+import 'dart:js_interop';
```
marandaneto, line 1: do we need to bump the min. version of dart and flutter here?
https://dart.dev/interop/js-interop/package-web#package-web-vs-dart-html
https://github.com/PostHog/posthog-flutter/blob/e44de573ea0aa2068d7c33a839d2c12130b1887c/pubspec.yaml#L10-L11

lib/src/posthog_flutter_web_handler.dart
```diff
@@ -1,107 +1,180 @@
...
+  external void opt_out_capturing();
+  external JSAny? getFeatureFlag(JSAny key);
+  external JSAny? getFeatureFlagPayload(JSAny key);
+  external void register(JSAny properties);
+  external void unregister(JSAny key);
+  external JSAny? get_session_id();
+}
+
+// Acessar o posthog a partir do window
+@JS('window.posthog')
```
marandaneto, line 31: if the script fails to set `window.posthog`, this will return null, so then `posthog` should be marked as nullable

lib/src/posthog_flutter_web_handler.dart
```diff
@@ -1,107 +1,180 @@
...
+JSAny stringToJSAny(String value) {
+  return value.toJS;
+}
+
+JSAny boolToJSAny(bool value) {
+  return value.toJS;
+}
+
+JSAny mapToJSAny(Map<dynamic, dynamic> map) {
+  return map.jsify()!;
```
marandaneto, line 44: fallback to an empty obj instead of forcing and crashing with `!`

lib/src/posthog_flutter_web_handler.dart
```diff
@@ -1,107 +1,180 @@
...
+      posthog.alias(
+        stringToJSAny(alias),
+      );
       break;
     case 'distinctId':
-      final distinctId = analytics.callMethod('get_distinct_id');
-
-      return distinctId;
+      final distinctId = posthog.get_distinct_id();
+      return distinctId?.dartify();
```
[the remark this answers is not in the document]
marandaneto: we didnt do this before but it was not type safety and now it is

pedrovmc:
Hey @marandaneto, thanks for the review! I'll probably go through it by the end of today or tomorrow morning. Thanks again!

lib/src/posthog_flutter_web_handler.dart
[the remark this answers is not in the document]
marandaneto: also if its empty, we should return null, similar to `getSessionId`

marandaneto:
> Hey @marandaneto, thanks for the review! I'll probably go through it by the end of today or tomorrow morning. Thanks again!

@pedrovmc nice, just finished the review and tested a few basic things, way to go, thanks and looking forward to your comments/commits.

pedrovmc:
## @marandaneto

Thank you for your comments and suggestions! I've made all the requested changes:

1. Removed the `web` package dependency since we're not directly using any imports from it (sorry about that, I was migrating an application that needed it, it's was almost automatic for me).
2. Adjusted minimum SDK requirements to Dart 3.3.0 and Flutter 3.19.0. (https://medium.com/dartlang/history-of-js-interop-in-dart-98b06991158f)
3. Fixed the `mapToJSAny` method to return `map.jsify() ?? JSObject()` instead of using the `!` operator
4. Improved type handling in the `distinctId` method to return `as String?`
5. Added verification to return `null` when `distinctId` is empty, similar to the `getSessionId` method

I've tested all these changes successfully in a WASM environment. The plugin is now more type-safe and fully compatible with the latest Flutter WASM builds.

Please let me know if you need any further adjustments.

Issue #81, opened 2024-02-13
Support `onFeatureFlags` callback


marandaneto:
### Description

https://posthog.com/docs/libraries/js#ensuring-flags-are-loaded-before-usage

- [x] Flutter
- [ ] RN
- [X] Android
- [X] iOS

marandaneto:
I think this could be possible with https://api.flutter.dev/flutter/services/MethodChannel/setMethodCallHandler.html
example

```dart
// Dart code
import 'package:flutter/services.dart';

const platform = MethodChannel('com.example.channel');

void setupNativeCallbacks() {
  platform.setMethodCallHandler((call) async {
    switch (call.method) {
      case 'onEvent':
        print('Received event from native: ${call.arguments}');
        // You can call any Dart function or class method here
        return 'Received';
      default:
        throw PlatformException(
          code: 'Unimplemented',
          details: 'Method ${call.method} not implemented.',
        );
    }
  });
}
```

then on native side (kotlin-android as an example):

```kotlin
val channel = MethodChannel(flutterEngine.dartExecutor.binaryMessenger, "com.example.channel")
channel.invokeMethod("onEvent", "Hello from Android!")
```

so when the android `onFeatureFlags` is called, we forward the call using `invokeMethod`, then the dart side listens to this and calls the flutter `onFeatureFlags`

PR #183, opened 2025-06-05
feat: add onFeatureFlags callback

laf-rge:
This change introduces an `onFeatureFlags` callback to the `Posthog` Flutter SDK. The primary motivation is to provide a reliable way for developers to know when feature flags have been loaded from the PostHog server, especially on the first application launch where flags might not be immediately available. This helps prevent race conditions and ensures that feature flag checks are made against up-to-date data.

This change addresses and closes issue #81.

The callback behaves as follows:
-   **Web:** Provides a list of active flag keys and a map of flag keys to their variant values.
-   **Mobile (Android/iOS):** Serves as a signal that the native PostHog SDK has finished loading feature flags. The `flags` and `flagVariants` parameters in the callback will be empty. Developers should use `Posthog().getFeatureFlag('key')` or `Posthog().isFeatureEnabled('key')` to retrieve specific flag values after this callback is invoked. This is because the native libraries do not have public methods for retrieving all features.

The changes were tested through several methods:
-   **Unit Tests:** Added new unit tests for the Dart `PosthogFlutterIO` implementation to cover callback registration, successful event handling (with and without data), error handling, and malformed data scenarios.
-   **Example App:** Ensured the example application in the repository compiles and builds successfully for both Android and iOS platforms after incorporating the native changes.
-   **Real-world Project Integration:** I tested this branch in a personal project that uses `posthog-flutter`. The dependency in `pubspec.yaml` was pointed to this forked version, and I confirmed that the `onFeatureFlags` callback fires correctly once the feature flags are available from the native SDKs.
-   **Native Code Verification:**
    -   For Android, verified that `channel.invokeMethod` is called on the main UI thread and that it signals readiness without attempting to send all flag data (as the native Android SDK's callback doesn't provide it directly).
    -   For iOS, confirmed that the plugin observes `PostHogSDK.didReceiveFeatureFlags` and, similar to Android, signals readiness without sending all flag data due to the absence of a public API in the native iOS SDK to fetch all flags and variants at once.

## :pencil: Checklist

- [x] I updated the docs if needed. (Updated Dartdoc for `onFeatureFlags` and added an entry to `CHANGELOG.md`)
- [x] No breaking change or entry added to the changelog. (New non-breaking feature; changelog entry added under "Next")

### DIFF
--- a/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
+++ b/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
@@ -30,8 +30,6 @@ class PosthogFlutterPlugin :
 
     private lateinit var applicationContext: Context
 
-    private val mainHandler = Handler(Looper.getMainLooper())
-
     private val snapshotSender = SnapshotSender()
 
     override fun onAttachedToEngine(flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
@@ -272,11 +270,11 @@ class PosthogFlutterPlugin :
                     try {
                         Log.i("PostHogFlutter", "Android onFeatureFlags triggered. Notifying Dart.")
                         val arguments = emptyMap<String, Any?>()
-                        mainHandler.post { channel.invokeMethod("onFeatureFlagsCallback", arguments) }
+                        channel.invokeMethod("onFeatureFlagsCallback", arguments)
                     } catch (e: Exception) {
                         Log.e("PostHogFlutter", "Error in onFeatureFlags signalling: ${e.message}", e)
                         val errorArguments = emptyMap<String, Any?>()
-                        mainHandler.post { channel.invokeMethod("onFeatureFlagsCallback", errorArguments) }
+                        channel.invokeMethod("onFeatureFlagsCallback", errorArguments)
                     }
                 }
             }


marandaneto:
@ioannisj would you like to review and test this one as part of owning a bit more of this SDK? happy to do a final review as well

android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -257,12 +262,36 @@ class PosthogFlutterPlugin :
                 posthogConfig.getIfNotNull<Boolean>("sessionReplay") {
                     sessionReplay = it
                 }
+                posthogConfig.getIfNotNull<String>("dataMode") {
+                    // Assuming DataMode is an enum or similar, handle appropriately
+                }
```
marandaneto, line 267: this is not used on android so we dont need it

android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -257,12 +262,36 @@ class PosthogFlutterPlugin :
...
+
+                onFeatureFlags = PostHogOnFeatureFlags {
+                    try {
+                        Log.i("PostHogFlutter", "Android onFeatureFlags triggered. Notifying Dart.")
+                        val arguments = mapOf(
+                            "flags" to emptyList<String>(),
+                            "flagVariants" to emptyMap<String, Any?>(),
+                            "errorsLoading" to false
+                        )
+                        mainHandler.post { channel.invokeMethod("onFeatureFlagsCallback", arguments) }
```
marandaneto, line 282: do we need to call this always within the main thread?

lib/src/posthog.dart
```diff
@@ -127,5 +127,52 @@ class Posthog {
...
+  ///   if (isNewFeatureEnabled) {
+  ///     // Implement logic for 'new-feature'
+  ///   }
+  ///   final variantValue = await Posthog().getFeatureFlag('multivariate-flag');
+  ///   if (variantValue == 'test-variant') {
+  ///     // Implement logic for 'test-variant' of 'multivariate-flag'
+  ///   }
+  /// });
+  /// ```
+  void onFeatureFlags(OnFeatureFlagsCallback callback) =>
```
marandaneto, line 174: Following the other SDKs, this should be a callback defined in the `PostHogConfig` during SDK init

marandaneto:
Oh I see this is similar to the Web implementation https://posthog.com/docs/libraries/js/features#ensuring-flags-are-loaded-before-usage hence the `featureFlags`, `flagVariants` and `errorsLoading` always empty.
Since the Mobile SDKs dont support this yet, I recommend removing them

marandaneto:
can we do this for the Web implementation as well?
https://github.com/PostHog/posthog-flutter/blob/main/lib/src/posthog_flutter_web_handler.dart

android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
laf-rge: That's a good question! We're using mainHandler.post here to ensure that channel.invokeMethod("onFeatureFlagsCallback", ...) is executed on the Android main thread.
The primary reason is that the Dart callback (onFeatureFlagsCallback) triggered by this method channel invocation might lead to UI updates or interact with Flutter framework elements that expect to be on the main isolate (which corresponds to the platform's main thread). 

I'm not experienced enough to know if we should do it on the main thread or not outside of that. Open to suggestions.

marandaneto: I think we should just use the current thread and Flutter itself when listening to `onFeatureFlagsCallback` should make sure it's the UI thread, IMO its the listener responsability

PR #188, opened 2025-06-27
feat: flutter surveys for iOS

ioannisj:

Docs PR: https://github.com/PostHog/posthog.com/pull/12209

This implements [PostHogSurveysDelegate](https://github.com/PostHog/posthog-ios/pull/359) for iOS to add surveys functionality to Flutter

## :pencil: Checklist

### DIFF
--- a/lib/src/posthog.dart
+++ b/lib/src/posthog.dart
@@ -119,6 +119,9 @@ class Posthog {
 
   Future<void> flush() => _posthog.flush();
 
+  /// Closes the PostHog SDK and cleans up resources.
+  ///
+  /// Note: Please note that after calling close(), surveys will not be rendered until the SDK is re-initialized and the next navigation event occurs.
   Future<void> close() {
     _config = null;
     _currentScreen = null;
--- a/lib/src/posthog_config.dart
+++ b/lib/src/posthog_config.dart
@@ -32,6 +32,9 @@ class PostHogConfig {
   var dataMode = PostHogDataMode.any;
 
   /// Enable Surveys (Currently for iOS only)
+  ///
+  /// Note: Please note that after calling Posthog().close(), surveys will not be rendered until the SDK is re-initialized and the next navigation event occurs.
+  ///
   /// Experimental. Defaults to false.
   @experimental
   var surveys = false;


ios/Classes/PostHogDisplaySurvey+Dict.swift
```diff
@@ -0,0 +1,104 @@
+#if os(iOS)
+    import Foundation
+    import PostHog
+
+    extension PostHogDisplaySurvey {
+        // Convert the survey object to a dictionary for communication with the Dart layer
+        // Native platform model -> Dictionary -> Dart model
+        func toDict() -> [String: Any] {
+            var dict: [String: Any] = [
+                "id": id,
```
marandaneto, line 10: not a blocker but we could use constants for the key names coming directly from the iOS SDK
or the iOS class would have a builtin .toDict that is reused by this plugin, so the code lives in the iOS SDK instead

lib/src/posthog_flutter_io.dart
```diff
@@ -1,17 +1,131 @@
...
 /// An implementation of [PosthogFlutterPlatformInterface] that uses method channels.
 class PosthogFlutterIO extends PosthogFlutterPlatformInterface {
+  PosthogFlutterIO() {
+    _methodChannel.setMethodCallHandler(_handleMethodCall);
+  }
+
   /// The method channel used to interact with the native platform.
   final _methodChannel = const MethodChannel('posthog_flutter');
 
+  Future<dynamic> _handleMethodCall(MethodCall call) async {
```
marandaneto, line 25: what will happen when we call those methods on Android and Web since its iOS only? we have to degrade gracefully with logging message instead

pubspec.yaml
```diff
@@ -15,6 +15,8 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
+  flutter_localizations:
+    sdk: flutter
```
marandaneto, line 19: do we still need this?

ios/Classes/PostHogDisplaySurvey+Dict.swift
marandaneto: that's correct, there will be duplication, but 2x better than 3x, your call here.

lib/posthog_flutter.dart
```diff
@@ -1,5 +1,3 @@
-library posthog_flutter;
```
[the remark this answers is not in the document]
marandaneto: I'd keep `library posthog_flutter;` because this could be not optional in older versions, so I'd rather fix or ignore the linter if it makes this line optional in newer versions

lib/src/posthog_flutter_io.dart
marandaneto: I'd run a test on android and web with surveys enabled just to make sure we are not crashing peoples apps but rather degrading gracefully

lib/src/posthog_observer.dart
```diff
@@ -20,6 +20,10 @@ class PosthogObserver extends RouteObserver<ModalRoute<dynamic>> {
       : _nameExtractor = nameExtractor,
         _routeFilter = routeFilter;
 
+  /// The current navigation context, which can be used for showing modals
+  /// This is updated whenever routes change (push, pop, replace)
+  static BuildContext? currentContext;
```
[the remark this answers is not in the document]
marandaneto: we should clear this var when the SDK is closed (called posthog.close())

lib/src/posthog_flutter_io.dart
```diff
@@ -1,17 +1,119 @@
...
+  PosthogFlutterIO() {
+    _methodChannel.setMethodCallHandler(_handleMethodCall);
+  }
+
   /// The method channel used to interact with the native platform.
   final _methodChannel = const MethodChannel('posthog_flutter');
 
+  /// Native plugin calls to Flutter
+  ///
+  Future<dynamic> _handleMethodCall(MethodCall call) async {
```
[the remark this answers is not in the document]
marandaneto: > I think it's the first time we are calling flutter/dart from native code, but I'll double check

makes sense, my bad

lib/src/posthog_flutter_io.dart
marandaneto: if its tested, sure

PR #212, opened 2025-10-16
feat: add manual error capture

ioannisj:
Part of https://github.com/PostHog/posthog/issues/38845
Needs https://github.com/PostHog/posthog/pull/39738

Docs: https://github.com/PostHog/posthog.com/pull/13356

CleanShot 2025-10-16 at 13 43 00

## :pencil: Checklist

### DIFF
--- a/lib/src/error_tracking/dart_exception_processor.dart
+++ b/lib/src/error_tracking/dart_exception_processor.dart
@@ -1,4 +1,3 @@
-import 'package:flutter/foundation.dart';
 import 'package:stack_trace/stack_trace.dart';
 import 'utils/isolate_utils.dart' as isolate_utils;
 
@@ -174,7 +173,7 @@ class DartExceptionProcessor {
     }
 
     // add function, if available
-    final member = _extractMember(frame);
+    final member = frame.member;
     if (member != null && member.isNotEmpty) {
       frameData['function'] = member;
     }
@@ -240,25 +239,6 @@ class DartExceptionProcessor {
 
     // 4. Default fallback
     return inAppByDefault;
-  }
-
-  static String? _extractMember(Frame frame) {
-    final member = frame.member;
-    if (member == null || member.isEmpty) {
-      return member;
-    }
-
-    // Only sanitize for web builds to avoid PII leakage
-    if (kIsWeb) {
-      // Handle WebAssembly format: "module0.InitialScreenState.build closure at file:///path/to/file.dart:251:30"
-      // For privacy, just return the function part (everything before "at")
-      final atIndex = member.indexOf(' at ');
-      if (atIndex != -1) {
-        return member.substring(0, atIndex);
-      }
-    }
-
-    return member;
   }
 
   static String? _extractPackage(Frame frame) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
     - Automatic exception capture
     - De-obfuscating stacktraces from obfuscated builds ([--obfuscate](https://docs.flutter.dev/deployment/obfuscate) and [--split-debug-info](https://docs.flutter.dev/deployment/obfuscate))
     - [Source code context](/docs/error-tracking/stack-traces) associated with an exception
+    - Flutter web support
   - **BREAKING**: Minimum Dart SDK version bumped to 3.4.0 and Flutter to 3.22.0 (required for `stack_trace` dependency compatibility)
 
 ## 5.6.0
--- a/lib/src/posthog_flutter_web_handler.dart
+++ b/lib/src/posthog_flutter_web_handler.dart
@@ -10,8 +10,7 @@ class PostHog {}
 extension PostHogExtension on PostHog {
   external JSAny? identify(
       JSAny userId, JSAny properties, JSAny propertiesSetOnce);
-  external JSAny? capture(
-      JSAny eventName, JSAny properties, CaptureOptions? options);
+  external JSAny? capture(JSAny eventName, JSAny properties);
   external JSAny? alias(JSAny alias);
   // ignore: non_constant_identifier_names
   external JSAny? get_distinct_id();
@@ -32,19 +31,6 @@ extension PostHogExtension on PostHog {
   external void unregister(JSAny key);
   // ignore: non_constant_identifier_names
   external JSAny? get_session_id();
-}
-
-@JS('Date')
-extension type JSDate._(JSObject _) implements JSObject {
-  external JSDate(num millisecondsSinceEpoch);
-}
-
-// PostHog capture options class
-@JS()
-@anonymous
-@staticInterop
-class CaptureOptions {
-  external factory CaptureOptions({JSDate? timestamp});
 }
 
 // Accessing PostHog from the window object
@@ -104,7 +90,6 @@ Future<dynamic> handleWebMethodCall(MethodCall call) async {
       posthog?.capture(
         stringToJSAny(eventName),
         mapToJSAny(properties),
-        null,
       );
       break;
     case 'screen':
@@ -115,7 +100,6 @@ Future<dynamic> handleWebMethodCall(MethodCall call) async {
       posthog?.capture(
         stringToJSAny('\$screen'),
         mapToJSAny(properties),
-        null,
       );
       break;
     case 'alias':
@@ -227,21 +211,7 @@ Future<dynamic> handleWebMethodCall(MethodCall call) async {
       // not supported on Web
       break;
     case 'captureException':
-      // grab event properties and timestamp from args
-      final properties = safeMapConversion(args['properties']);
-      final timestamp = args['timestamp'] as int?;
-
-      // The properties already contain the processed exception data from DartExceptionProcessor.processException
-      if (timestamp != null) {
-        final options = CaptureOptions(timestamp: JSDate(timestamp));
-
-        posthog?.capture(
-            stringToJSAny('\$exception'), mapToJSAny(properties), options);
-      } else {
-        // No timestamp provided: skip capture to avoid out-of-order events
-        print(
-            'PostHog: Skipping exception capture without timestamp to maintain event order');
-      }
+      // not implemented on Web
       break;
     default:
       throw PlatformException(


android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -532,6 +535,20 @@ class PosthogFlutterPlugin :
...
+    private fun captureException(call: MethodCall, result: Result) {
+        try {
+            val arguments = call.arguments as? Map<String, Any> ?: run {
+                result.error("INVALID_ARGUMENTS", "Invalid arguments for captureException", null)
+                return
+            }
+            
+            PostHog.capture("\$exception", properties = arguments)
+            result.success(null)
+        } catch (e: Exception) {
```
marandaneto:
```suggestion
        } catch (e: Throwable) {
```

lib/src/posthog_flutter_io.dart
```diff
@@ -413,6 +420,34 @@ class PosthogFlutterIO extends PosthogFlutterPlatformInterface {
...
+    bool handled = true,
+  }) async {
+    if (!isSupportedPlatform()) {
+      return;
+    }
+
+    try {
+      final exceptionData = DartExceptionProcessor.processException(
+        error: error,
+        stackTrace: stackTrace ?? StackTrace.current,
```
ablaszkiewicz: Same here with the `StackTrace.current`

lib/src/exceptions/dart_exception_processor.dart
```diff
@@ -0,0 +1,293 @@
+import 'package:stack_trace/stack_trace.dart';
+import 'utils/isolate_utils.dart' as isolate_utils;
+
+class DartExceptionProcessor {
```
marandaneto: lets rename the folder to errortracking instead of exceptions

ablaszkiewicz (dismissed):
Never seen dart before but looks legit

lib/src/exceptions/dart_exception_processor.dart
```diff
@@ -0,0 +1,293 @@
...
+    Frame frame, {
+    List<String>? inAppIncludes,
+    List<String>? inAppExcludes,
+    bool inAppByDefault = true,
+  }) {
+    final member = frame.member;
+    final fileName = _extractFileName(frame);
+
+    final frameData = <String, dynamic>{
+      'function': member ?? 'unknown',
```
marandaneto: i'd rather make it optional in the frontend and backend instead of fallback sdk side

lib/src/exceptions/dart_exception_processor.dart
```diff
@@ -0,0 +1,293 @@
...
+      'module': _extractModule(frame),
+      'package': _extractPackage(frame),
```
marandaneto: check if theres a package before setting it

lib/src/posthog_config.dart
```diff
@@ -100,3 +104,50 @@ class PostHogSessionReplayConfig {
...
+  ///
+  /// inApp Example:
+  /// inAppIncludes = ["package:your_app", "package:your_company_utils"]
+  /// All exception stacktrace frames from these packages will be considered inApp
+  ///
+  /// This option takes precedence over inAppExcludes.
+  /// For Flutter/Dart, this typically includes:
+  /// - Your app's main package (e.g., "package:your_app")
+  /// - Any internal packages you own (e.g., "package:your_company_utils")
+  var inAppIncludes = <String>[];
```
marandaneto: i think we should make them final but mutable

android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -532,6 +535,20 @@ class PosthogFlutterPlugin :
...
     }
 
+    private fun captureException(call: MethodCall, result: Result) {
+        try {
+            val arguments = call.arguments as? Map<String, Any> ?: run {
+                result.error("INVALID_ARGUMENTS", "Invalid arguments for captureException", null)
+                return
+            }
+            
+            PostHog.capture("\$exception", properties = arguments)
```
marandaneto: we can make the android enum with event names public if not yet and reuse here (same for iOS)
the other thing is that this is an async call in flutter, and right now the event timestamp is created by the native sdk
the timestamp should come from the flutter sdk
so we will need a way to read and reuse the timestamp from the event before reading the current timestamp in the capture method

lib/src/posthog.dart
```diff
@@ -121,6 +121,25 @@ class Posthog {
...
+  /// [handled] - Whether the exception was handled (true by default for manual captures)
+  Future<void> captureException({
+    required dynamic error,
+    StackTrace? stackTrace,
+    Map<String, Object>? properties,
+    bool handled = true,
+  }) =>
+      _posthog.captureException(
+        error: error,
+        stackTrace: stackTrace ?? StackTrace.current,
```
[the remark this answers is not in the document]
marandaneto: and before doing this, you should check if its an error type
```
if (throwable is Error) {
      stackTrace ??= throwable.stackTrace;
}
```

so you use the stacktrace from the instance instead the current so it should not be read in the method param directly

lib/src/exceptions/dart_exception_processor.dart
```diff
@@ -0,0 +1,293 @@
...
+  static Map<String, dynamic> processException({
+    required dynamic error,
+    required StackTrace stackTrace,
+    Map<String, Object>? properties,
+    bool handled = true,
+    List<String>? inAppIncludes,
+    List<String>? inAppExcludes,
+    bool inAppByDefault = true,
+  }) {
+    // Process single exception (Dart doesn't provide standard exception chaining afaik)
```
marandaneto: errors could have a cause, but the dart exception does not have the cause attribute in it
we'd need a ExceptionCause approach where customers could read and return the cause per exception
we dont need this in v1 just keep in mind
quick example:
```
/// class ExceptionWithInner {
///   ExceptionWithInner(this.innerException, this.innerStackTrace);
///   Object innerException;
///   StackTrace innerStackTrace;
/// }
///
/// class ExceptionWithInnerExtractor extends ExceptionCauseExtractor<ExceptionWithInner>  {
///   @override
///   ExceptionCause? cause(ExceptionWithInner error) {
///     return ExceptionCause(error.innerException, error.innerStackTrace);
///   }
/// }
///
/// options.addExceptionCauseExtractor(ExceptionWithInnerExtractor());
```

lib/src/exceptions/utils/isolate_utils.dart
```diff
@@ -0,0 +1,10 @@
+import '_io_isolate_utils.dart'
+    if (dart.library.js_interop) '_web_isolate_utils.dart' as platform;
```
marandaneto: used to be dart.library.html or io here for web vs io
just make sure `js_interop` is available since the min version we support for our sdk

lib/src/posthog_config.dart
ioannisj: Do you mean something like:

`config.inAppIncludes.add('package:my_app');` 
instead of 
`config.inAppIncludes = ['package:my_app'];`?

Wonder what's the benefit of that?

lib/src/error_tracking/dart_exception_processor.dart
```diff
@@ -0,0 +1,293 @@
...
+    } catch (e) {
+      // Graceful fallback if isolate detection fails
+      return 1;
+    }
+  }
+
+  static String _getExceptionType(dynamic error) {
+    // For primitives (String, int, bool, double, null, etc.), just use "Error"
+    if (_isPrimitive(error)) {
+      return 'Error';
```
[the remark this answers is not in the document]
marandaneto: I actually prefer to keep the original type, its much more info (and useful) than a generic `Error`, you can do this by just reading its runtimeType

lib/src/posthog_flutter_io.dart
```diff
@@ -413,6 +420,34 @@ class PosthogFlutterIO extends PosthogFlutterPlatformInterface {
...
+        error: error,
+        stackTrace: stackTrace,
+        properties: properties,
+        handled: handled,
+        inAppIncludes: _config?.errorTrackingConfig.inAppIncludes,
+        inAppExcludes: _config?.errorTrackingConfig.inAppExcludes,
+        inAppByDefault: _config?.errorTrackingConfig.inAppByDefault ?? true,
+      );
+
+      await _methodChannel.invokeMethod('captureException', exceptionData);
```
marandaneto: before we call the method, we need to be sure that `exceptionData` is serializable, since `properties` is `Map<String, Object>`, we cannot trust the values, we do something similar for Android and iOS, we'd need to normalize the data here too otherwise this method will fail

lib/src/error_tracking/dart_exception_processor.dart
```diff
@@ -0,0 +1,335 @@
...
+      '\$exception_list': [exceptionData],
+    };
+
+    // Add custom properties if provided
+    if (properties != null) {
+      for (final entry in properties.entries) {
+        // Don't allow overwriting system properties
+        if (!result.containsKey(entry.key)) {
+          result[entry.key] = entry.value;
+        }
```
marandaneto: I'd say the given props always have precedence, like we do for all analytics events etc, so just merge the given props last, and if users set something internal/system props, that was their intention, similar to the beforeSend callback, they can change anything

lib/src/error_tracking/dart_exception_processor.dart
```diff
@@ -0,0 +1,335 @@
...
+    final chain = Chain.forTrace(stackTrace);
+    final frames = <Map<String, dynamic>>[];
+
+    for (final trace in chain.traces) {
+      bool skipNextPostHogFrame = removeTopPostHogFrames;
+
+      for (final frame in trace.frames) {
+        // Skip top PostHog frames?
+        if (skipNextPostHogFrame) {
+          if (_isPostHogFrame(frame)) {
```
marandaneto: we should only skip if we generated the stack trace, otherwise it should be there because it was likely a bug on our end or something

lib/src/error_tracking/dart_exception_processor.dart
```diff
@@ -0,0 +1,335 @@
...
+    // For primitives (String, int, bool, double, null, etc.), just use "Error"
+    if (_isPrimitive(error)) {
+      return 'Error';
+    }
+
+    return error.runtimeType.toString();
+  }
+
+  /// Checks if a value is a primitive type
+  static bool _isPrimitive(dynamic value) {
```
marandaneto: i'd avoid the use of `dynamic` everywhere, always prefer Object, so we know if its nullable or not

pubspec.yaml
```diff
@@ -18,6 +18,7 @@ dependencies:
   plugin_platform_interface: ^2.0.2
   # plugin_platform_interface depends on meta anyway
   meta: ^1.3.0
+  stack_trace: ^1.11.1
```
marandaneto: see https://github.com/PostHog/posthog-flutter/blob/b339e95e29c2f152f03a8581c101a98da6f175f4/pubspec.yaml#L10
our min dart version is currently 3.3
v1.11.1 supports Dart >= 2.18 https://pub.dev/packages/stack_trace/versions
so we'd need to use min [1.12.0](https://pub.dev/packages/stack_trace/versions/1.12.0) and bump our min version of the Dart SDK and Flutter to 3.4.0 as well to be compatible with the stack trace version

lib/src/error_tracking/dart_exception_processor.dart
```diff
@@ -0,0 +1,335 @@
...
+
+    // If it's an Error, try to use its built-in stackTrace
+    if (error is Error) {
+      effectiveStackTrace ??= error.stackTrace;
+    }
+
+    // If still null or empty, get current stack trace
+    if (effectiveStackTrace == null ||
+        effectiveStackTrace == StackTrace.empty) {
+      effectiveStackTrace = stackTraceProvider?.call() ?? StackTrace.current;
```
marandaneto: you have to check again for StackTrace.empty btw since `StackTrace.current` could return empty

lib/src/error_tracking/dart_exception_processor.dart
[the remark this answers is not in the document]
ioannisj: yeah my only concern is keeping behavior aligned with the js sdk, but i’m fine dropping the primitive check and just returning the type 

personally i don’t think "String", "int", "bool" adds much context tbh, but no strong feelings either way. Was more of a UI concern at the time, which can be handled in FE anw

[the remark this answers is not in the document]
marandaneto: > yeah my only concern is keeping behavior aligned with the js sdk

different languages and platforms have different needs, so we dont have to align unless its just the same.
we should focus on the best UX of each language/platforn
i recommend you checking how the stacktrace looks like in the console when an exception is thrown (let the language/IDE print the stacktrace), we should use that as a template which is the one that users are used to read and fix things

> personally i don’t think "String", "int", "bool" adds much context tbh, but no strong feelings either way

if you are debugging an issue and you see 'String thrown' vs 'Error thrown', would you agree that is a bit more signal?
If you are grouping errors, would you agree that 2 errors that are Strings should be grouped together than just all thrown primitive types together?
I think the 2 cases above make it a better signal than a generic error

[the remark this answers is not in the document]
ioannisj: >i recommend you checking how the stacktrace looks like in the console when an exception is thrown (let the language/IDE print the stacktrace), we should use that as a template which is the one that users are used to read and fix things

It just prints the string representation of the value, no type info.

> if you are debugging an issue and you see 'String thrown' vs 'Error thrown', would you agree that is a bit more signal?
If you are grouping errors, would you agree that 2 errors that are Strings should be grouped together than just all thrown primitive types together?
I think the 2 cases above make it a better signal than a generic error

Yeah, fair points. Let's not overthink it then. Let's go with always `e.runtimeType.toString()`

lib/src/error_tracking/utils/_io_isolate_utils.dart
```diff
@@ -0,0 +1,16 @@
...
+/// Gets the current isolate's debug name for IO platforms
+String? getIsolateName() => Isolate.current.debugName;
+
+/// Determines if the current isolate is the root isolate for IO platforms
+/// Uses Flutter's ServicesBinding to detect the root isolate
+bool isRootIsolate() {
+  try {
+    return ServicesBinding.rootIsolateToken != null;
+  } catch (_) {
+    // If ServicesBinding is not available (pure Dart), assume root isolate
```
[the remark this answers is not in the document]
ioannisj: Yeah, you are right. As far as I know ServicesBinding.rootIsolateToken is guaranteed to exists in IO platforms. So we may not even need the try-catch here, even though doesn't hurt to keep either. Web always returns true here

lib/src/error_tracking/utils/isolate_utils.dart
```diff
@@ -0,0 +1,10 @@
+import '_io_isolate_utils.dart'
+    if (dart.library.js_interop) '_web_isolate_utils.dart' as platform;
```
ioannisj: I see references in [Dart 3.1.0](https://dart.googlesource.com/sdk/+/stable/CHANGELOG.md#3_1_0-2023_08_16) changelog, so I think we are okay here. If there's a better way to check, do let me know though

android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
marandaneto: the name of the event sure, the timestamp is important otherwise you'd be releasing something that might be off (out of the timeline)

ioannisj: For iOS there's already a capture() overload that accepts a timestamp, so makes sense to use that. If I remember correctly Android doesn't have that yet so we'll need to add it

ioannisj: Either way, won't release until timestamps come at the time of capture from Flutter side

example/lib/main.dart
```diff
@@ -21,6 +22,28 @@ Future<void> main() async {
...
+      'errorType': e.runtimeType.toString(),
+      'stackTrace': stack.toString(),
+      'properties': {
+        'test_type': 'background_isolate_exception',
+        'isolate_name': 'exception-demo-worker',
+        'button_pressed': 'capture_exception_background',
+      },
+    });
+  }
+}
```
marandaneto: i still find this example odd, this is not something an user would do.
Exceptions will bubble up to the main isolate anyway
whats your intention here?

lib/src/utils/property_normalizer.dart
```diff
@@ -0,0 +1,26 @@
+class PropertyNormalizer {
+  /// Normalizes a map of properties to ensure they are serializable through method channels.
+  ///
+  /// Unsupported types are converted to strings using toString().
+  /// Nested maps and lists are recursively normalized.
+  static Map<String, dynamic> normalize(Map<dynamic, dynamic> properties) {
```
marandaneto: I'd not use dynamic only Object

lib/src/utils/property_normalizer.dart
```diff
@@ -0,0 +1,26 @@
...
+    if (value == null) {
+      return null;
+    }
+    if (value is bool || value is String || value is double || value is int) {
+      return value;
+    } else if (value is List<dynamic>) {
+      return value.map((e) => _normalizeValue(e)).toList();
+    } else if (value is Map) {
+      return normalize(value);
+    } else {
```
marandaneto: do we need to add `Set` here as well?
https://dart.dev/language/built-in-types

lib/src/posthog.dart
```diff
@@ -121,6 +121,22 @@ class Posthog {
...
+  /// [properties] - Optional custom properties to attach to the exception event
+  Future<void> captureException(
+          {required Object error,
+          StackTrace? stackTrace,
+          Map<String, Object>? properties}) =>
+      _posthog.captureException(
+        error: error,
+        stackTrace: stackTrace,
+        properties: properties,
+        handled: true,
```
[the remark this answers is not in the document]
ioannisj: I think you are right. This is on PosthogFlutterPlatformInterface which is used internally in main Posthog class, but the type is still public I think. Not sure if users would interact with this directory but better pass this as a property anw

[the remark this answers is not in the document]
marandaneto: no need to use a prop, check the android SDK, we use an exception wrapper to find out if its handled or not handled.
in the future we would have a param that has more info than just a boolean as mentioned before

lib/src/utils/property_normalizer.dart
```diff
@@ -0,0 +1,26 @@
...
+  static dynamic _normalizeValue(dynamic value) {
+    if (value == null) {
+      return null;
+    }
+    if (value is bool || value is String || value is double || value is int) {
+      return value;
+    } else if (value is List<dynamic>) {
+      return value.map((e) => _normalizeValue(e)).toList();
+    } else if (value is Map) {
+      return normalize(value);
```
marandaneto: here you need to do something similar to `List`, normalize the key and value and toMap at the end

lib/src/utils/property_normalizer.dart
marandaneto: unless the Set is a subtype of List, we will stringify the set which isnt ideal, we can make a list of it, and normalize its values ofc

ioannisj:
> @ioannisj asked on the official docs but worth checking, do we need a guardrail for flutter web?

Replied there https://github.com/PostHog/posthog.com/pull/13356#discussion_r2455489061. Pushed a commit and now captureException is a noop on web

Question is, should we add support now? I'd have to see how this will play with js config options. We have the event data and props, so we may as well capture the event. My take is focus on mobile for now, then come back to web? wdyt?

lib/src/utils/property_normalizer.dart
```diff
@@ -1,26 +1,47 @@
...
-    if (value is bool || value is String || value is double || value is int) {
+  static Object? _normalizeValue(Object? value) {
+    if (_isSupported(value)) {
       return value;
-    } else if (value is List<dynamic>) {
+    } else if (value is List) {
       return value.map((e) => _normalizeValue(e)).toList();
     } else if (value is Map) {
-      return normalize(value);
+      return Map<String, Object?>.fromEntries(
```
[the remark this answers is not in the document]
ioannisj: Good question, idk for sure. I'd think it's okay because of the more permissive target type. Switch to direct iteration just to be safe here

ioannisj:
> @ioannisj asked on the official docs but worth checking, do we need a guardrail for flutter web?

Okay took a while but validated that web runs okay so worth including. Had a rough time with js interop but managed to figure it out thanx to [this](https://github.com/dart-lang/site-www/blob/c6e3c8bf7cb7589cd6b656ea21ade8234cefd424/src/content/interop/js-interop/usage.md?plain=1#L359)

Tested for default as well [WebAssembly](https://docs.flutter.dev/platform-integration/web/renderers) build modes. 

When building with --wasm, I saw that some function names were leaking some PII. They had the shape of `module0.InitialScreenState.build closure at file:///path/to/file.dart:251:30`. Tried extracting just function name part in those cases. It might not catch every shape, but we can iterate once we see real world reports.

### --no-wasm
CleanShot 2025-10-24 at 21 51 40@2x

### --wasm
CleanShot 2025-10-24 at 23 11 09@2x

### console
CleanShot 2025-10-24 at 23 11 44

lib/src/error_tracking/dart_exception_processor.dart
```diff
@@ -225,6 +226,25 @@ class DartExceptionProcessor {
     return inAppByDefault;
   }
 
+  static String? _extractMember(Frame frame) {
+    final member = frame.member;
+    if (member == null || member.isEmpty) {
+      return member;
+    }
+
+    // Only sanitize for web builds to avoid PII leakage
```
marandaneto: what type of PII?

lib/src/error_tracking/dart_exception_processor.dart
```diff
@@ -0,0 +1,295 @@
...
+
+  /// Parses stack trace into PostHog format
+  static List<Map<String, dynamic>> _parseStackTrace(
+    StackTrace stackTrace, {
+    List<String>? inAppIncludes,
+    List<String>? inAppExcludes,
+    bool inAppByDefault = true,
+    bool removeTopPostHogFrames = false,
+  }) {
+    final chain = Chain.forTrace(stackTrace);
```
marandaneto: check this out https://github.com/getsentry/sentry-dart/blob/a69a51fd1695dd93024be80a50ad05dd990b2b82/packages/dart/lib/src/sentry_stack_trace_factory.dart#L65C23-L69
stacktrace can be different types, so worth checking before doing `Chain.forTrace`

lib/src/error_tracking/dart_exception_processor.dart
```diff
@@ -225,6 +226,25 @@ class DartExceptionProcessor {
...
+  static String? _extractMember(Frame frame) {
+    final member = frame.member;
+    if (member == null || member.isEmpty) {
+      return member;
+    }
+
+    // Only sanitize for web builds to avoid PII leakage
+    if (kIsWeb) {
+      // Handle WebAssembly format: "module0.InitialScreenState.build closure at file:///path/to/file.dart:251:30"
+      // For privacy, just return the function part (everything before "at")
```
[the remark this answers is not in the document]
marandaneto: https://docs.flutter.dev/deployment/web#minification
looks like its minified, so prob not useful and we should disable web support for now (pls confirm thats the case)

lib/src/error_tracking/dart_exception_processor.dart
ioannisj: That was user paths but based on your next comments I'll see if release is indeed minified and worth the trouble.

lib/src/error_tracking/dart_exception_processor.dart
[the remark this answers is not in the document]
ioannisj: yeah confirmed, release builds are minified. Stacktraces aren't super helpful rn, but we still get the exception/error info which is the good win by itself. I think having *some* signal is better than zero signal, so I would ship as is. wdyt?

lib/src/error_tracking/dart_exception_processor.dart
marandaneto: There was a reason why it was done that way i just dont remember by heart, git blame might give you some hints

Issue #155, opened 2025-01-23
before_send support


marandaneto:
### Problem Statement

https://posthog.com/docs/libraries/js#sampling-events-using-a-custom-function

so you can sample/redact events

- [ ] Flutter
- [ ] iOS (already has propertiesSanitizer)
- [ ] RN
- [ ] Android (already has propertiesSanitizer)

PR #214, opened 2025-10-17
feat: autocapture unhandled exceptions

ioannisj:
Part of https://github.com/PostHog/posthog/issues/38845

Docs: https://github.com/PostHog/posthog.com/pull/13539

Adds autocapture for Flutter framework and Dart platform unhandled exceptions

- Unit tests: TODO

## :pencil: Checklist

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:flutter/material.dart';
 import 'package:posthog_flutter/posthog_flutter.dart';
 
@@ -22,6 +24,8 @@ Future<void> main() async {
       true; // Capture Flutter framework errors
   config.errorTrackingConfig.capturePlatformDispatcherErrors =
       true; // Capture Dart runtime errors
+  config.errorTrackingConfig.captureIsolateErrors =
+      true; // Capture isolate errors
 
   await Posthog().setup(config);
 
@@ -250,7 +254,7 @@ class InitialScreenState extends State<InitialScreen> {
                 const Padding(
                   padding: EdgeInsets.all(8.0),
                   child: Text(
-                    "Error Tracking",
+                    "Error Tracking - Manual",
                     style: TextStyle(fontWeight: FontWeight.bold),
                   ),
                 ),
@@ -305,6 +309,98 @@ class InitialScreenState extends State<InitialScreen> {
                     );
                   },
                   child: const Text("Capture Exception (Missing Stack)"),
+                ),
+                const Divider(),
+                const Padding(
+                  padding: EdgeInsets.all(8.0),
+                  child: Text(
+                    "Error Tracking - Autocapture",
+                    style: TextStyle(fontWeight: FontWeight.bold),
+                  ),
+                ),
+                ElevatedButton(
+                  style: ElevatedButton.styleFrom(
+                    backgroundColor: Colors.red,
+                    foregroundColor: Colors.white,
+                  ),
+                  onPressed: () {
+                    if (mounted) {
+                      ScaffoldMessenger.of(context).showSnackBar(
+                        const SnackBar(
+                          content:
+                              Text('Flutter error triggered! Check PostHog.'),
+                          backgroundColor: Colors.red,
+                          duration: Duration(seconds: 3),
+                        ),
+                      );
+                    }
+
+                    // Test Flutter error handler by throwing in widget context
+                    throw const CustomException(
+                        'Test Flutter error for autocapture',
+                        code: 'FlutterErrorTest',
+                        additionalData: {'test_type': 'flutter_error'});
+                  },
+                  child: const Text("Test Flutter Error Handler"),
+                ),
+                ElevatedButton(
+                  style: ElevatedButton.styleFrom(
+                    backgroundColor: Colors.blue,
+                    foregroundColor: Colors.white,
+                  ),
+                  onPressed: () {
+                    // Test PlatformDispatcher error handler with Future
+                    Future.delayed(Duration.zero, () {
+                      throw const CustomException(
+                          'Test PlatformDispatcher error for autocapture',
+                          code: 'PlatformDispatcherTest',
+                          additionalData: {
+                            'test_type': 'platform_dispatcher_error'
+                          });
+                    });
+
+                    if (mounted) {
+                      ScaffoldMessenger.of(context).showSnackBar(
+                        const SnackBar(
+                          content: Text(
+                              'Dart runtime error triggered! Check PostHog.'),
+                          backgroundColor: Colors.blue,
+                          duration: Duration(seconds: 3),
+                        ),
+                      );
+                    }
+                  },
+                  child: const Text("Test Dart Error Handler"),
+                ),
+                ElevatedButton(
+                  style: ElevatedButton.styleFrom(
+                    backgroundColor: Colors.purple,
+                    foregroundColor: Colors.white,
+                  ),
+                  onPressed: () {
+                    // Test isolate error listener by throwing in an async callback
+                    Timer(Duration.zero, () {
+                      throw const CustomException(
+                        'Isolate error for testing',
+                        code: 'IsolateHandlerTest',
+                        additionalData: {
+                          'test_type': 'isolate_error_listener_timer',
+                        },
+                      );
+                    });
+
+                    if (mounted) {
+                      ScaffoldMessenger.of(context).showSnackBar(
+                        const SnackBar(
+                          content:
+                              Text('Isolate error triggered! Check PostHog.'),
+                          backgroundColor: Colors.purple,
+                          duration: Duration(seconds: 3),
+                        ),
+                      );
+                    }
+                  },
+                  child: const Text("Test Isolate Error Handler"),
                 ),
                 const Divider(),
                 const Padding(


ioannisj:
Still need to address some feedback on https://github.com/PostHog/posthog-flutter/pull/212, so just looking for a general review on the autocapture approach/config for now 🙏

lib/src/posthog_config.dart
```diff
@@ -143,11 +143,58 @@ class PostHogErrorTrackingConfig {
...
+  /// shouldCaptureException: (error) {
+  ///   // Don't capture test exceptions
+  ///   if (error is TestException) return false;
+  ///   // Don't capture StateError exceptions
+  ///   if (error is StateError) return false;
+  ///   // Capture all other exceptions
+  ///   return true;
+  /// }
+  /// ```
+  bool Function(dynamic error)? shouldCaptureException;
```
marandaneto: we should implement https://github.com/PostHog/posthog-flutter/issues/155 instead of having another callback for the very same thing

lib/src/posthog_config.dart
```diff
@@ -143,11 +143,58 @@ class PostHogErrorTrackingConfig {
   /// - All other packages are inApp unless in inAppExcludes
   var inAppByDefault = true;
 
+  /// Enable automatic capture of unhandled exceptions
```
marandaneto: we could add one more config here for android autocapture https://github.com/PostHog/posthog-android/blob/3b565baef8b46510ca515a13f211e245edf1b4c6/posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt#L16

lib/src/exceptions/posthog_error_tracking_autocapture_integration.dart
```diff
@@ -0,0 +1,175 @@
...
+      return null;
+    }
+
+    _instance = PostHogErrorTrackingAutoCaptureIntegration._(
+      config: config,
+      posthog: posthog,
+    );
+
+    if (config.captureUnhandledExceptions) {
+      _instance!.start();
```
marandaneto: avoid the use of `!`

lib/src/exceptions/posthog_error_tracking_autocapture_integration.dart
```diff
@@ -0,0 +1,175 @@
...
+    // Restore original handlers
+    if (_originalFlutterErrorHandler != null) {
+      FlutterError.onError = _originalFlutterErrorHandler;
+      _originalFlutterErrorHandler = null;
+    }
+
+    if (_originalPlatformErrorHandler != null) {
+      PlatformDispatcher.instance.onError = _originalPlatformErrorHandler;
+      _originalPlatformErrorHandler = null;
+    }
```
marandaneto: we should restore to the default value regardless if they are null or not
https://github.com/PostHog/posthog-android/blob/3b565baef8b46510ca515a13f211e245edf1b4c6/posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt#L65
otherwise, if `_originalPlatformErrorHandler` is null, we will never unhook our own handlers

lib/src/exceptions/posthog_error_tracking_autocapture_integration.dart
```diff
@@ -0,0 +1,175 @@
...
+
+    _originalFlutterErrorHandler = FlutterError.onError;
+
+    FlutterError.onError = _posthogFlutterErrorHandler;
+  }
+
+  void _posthogFlutterErrorHandler(FlutterErrorDetails details) {
+    // don't capture silent errors (could maybe be a config?)
+    if (!details.silent) {
+      _captureException(
```
marandaneto: a few more details we should capture eg
https://github.com/getsentry/sentry-dart/blob/a69a51fd1695dd93024be80a50ad05dd990b2b82/packages/flutter/lib/src/integrations/flutter_error_integration.dart#L35-L60

lib/src/exceptions/posthog_error_tracking_autocapture_integration.dart
```diff
@@ -0,0 +1,175 @@
...
+    } else {
+      // If no original handler, use the default behavior (default is to dump to console)
+      FlutterError.presentError(details);
+    }
+  }
+
+  /// Platform error handler for Dart runtime errors
+  void _setupPlatformErrorHandler() {
+    // prevent circular calls
+    if (PlatformDispatcher.instance.onError == _posthogPlatformErrorHandler) {
```
marandaneto: this was added not a long ago so i am not sure this is available from our min version which is now 3.22
if this was added after 3.22, we have 2 options
either increase the min version or do this https://github.com/getsentry/sentry-dart/blob/a69a51fd1695dd93024be80a50ad05dd990b2b82/packages/flutter/lib/src/utils/platform_dispatcher_wrapper.dart#L49
which is figure out at runtime if its available or not and only set the callback if its available at runtime

marandaneto: iirc PlatformDispatcher.instance.onError does not work on flutter web, so it should be a no op
https://github.com/flutter/flutter/issues/100277

marandaneto: the alternative for web and older versions before having `PlatformDispatcher.instance.onError` is https://api.flutter.dev/flutter/dart-async/runZonedGuarded.html
https://github.com/getsentry/sentry-dart/blob/a69a51fd1695dd93024be80a50ad05dd990b2b82/packages/dart/lib/src/sentry_run_zoned_guarded.dart#L12
i'd avoid using `runZonedGuarded` if possible and only support the other things going forward

lib/src/exceptions/posthog_error_tracking_autocapture_integration.dart
```diff
@@ -0,0 +1,175 @@
...
+
+  Future<void> _captureException({
+    required dynamic error,
+    required StackTrace? stackTrace,
+    required bool handled,
+    String? context,
+  }) {
+    return _posthog.captureException(
+      error: error,
+      stackTrace: stackTrace ?? StackTrace.current,
```
marandaneto: do not use `StackTrace.current` here, you need to figure this out inside of the captureException so you can set syntheetic or not

lib/src/posthog_config.dart
ioannisj: Yeah agreed. This was more off your comment in one of our calls that it would be good idea to check directly on error type before capturing an exception. Will remove

lib/src/posthog_config.dart
```diff
@@ -143,11 +143,58 @@ class PostHogErrorTrackingConfig {
...
+  ///
+  /// When enabled, PostHog will automatically capture:
+  /// - Flutter framework errors (FlutterError.onError)
+  /// - Dart runtime errors (PlatformDispatcher.onError)
+  ///
+  /// Default: false
+  var captureUnhandledExceptions = false;
```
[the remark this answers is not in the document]
jakelacey2012: Some feedback as I'm really excited to use this feature!
- In the [elixir package](https://posthog.com/docs/libraries/elixir#error-tracking) the default behaviour is enabled, as a user I'd expect this to be consistent across libraries, but also means for projects like mine I don't have to think about configuring it, it should just work when I update.
- Really when I get started all I care about is capturing uncaught errors, and if its enabled by default then all I really need to think about is disabling it for development, or filtering out errors for certain things - those two granular things can be solved by the user if they wish with the shouldCaptureException function.

example/lib/main.dart
```diff
@@ -16,11 +17,50 @@ Future<void> main() async {
...
   config.sessionReplayConfig.throttleDelay = const Duration(milliseconds: 1000);
   config.flushAt = 1;
+
+  // Configure error tracking and exception capture
+  config.errorTrackingConfig.captureUnhandledExceptions =
+      true; // Enable autocapture
+  config.errorTrackingConfig.captureFlutterErrors =
+      true; // Capture Flutter framework errors
+  config.errorTrackingConfig.captureDartErrors =
+      true; // Capture Dart runtime errors
```
jakelacey2012: How will users configure this via the AndriodManifest & info.plist?

ioannisj: @jakelacey2012 Hey, not sure what you mean here? Maybe I'm missing something from the implementation. Mind elaborating?

jakelacey2012: > Mind elaborating?

@marandaneto beat me to it, but absolutely - there are two ways to configure the [Flutter SDK](https://posthog.com/docs/libraries/flutter) either automatically (via the AndriodManifest and info.plist) or manually within the dart code itself (at runtime). In this example you are illustrating how it will work for the manual setup, but customers (like myself) chose the easy way to go with the automatic setup.

> error tracking will only be supported if you do the 'manual init' so you dont need anything in the AndriodManifest and info.plist

Why is that? Just for my understanding is it just impractical? This decision means I now need to go from an automatic setup to a manual setup - which isn't the end of the world, but is friction never the less.

lib/src/error_tracking/posthog_error_tracking_autocapture_integration.dart
```diff
@@ -0,0 +1,175 @@
...
+    } else {
+      // If no original handler, use the default behavior (default is to dump to console)
+      FlutterError.presentError(details);
+    }
+  }
+
+  /// Platform error handler for Dart runtime errors
+  void _setupPlatformErrorHandler() {
+    // prevent circular calls
+    if (PlatformDispatcher.instance.onError == _posthogPlatformErrorHandler) {
```
ioannisj: Yeah came to the same conclusion, handler is not called. But I read the issue above more like a bug, i.e something that will eventually work. In that case shouldn't we just leave it? Also, based on our discussion in https://github.com/PostHog/posthog-flutter/pull/212#discussion_r2462766199, we skipped adding supported either way for Flutter web currently. So even if the handler is called, we'll skip capture.  I can just wrap this in `kIsWeb` just to be safe though

marandaneto: > But I read the issue above more like a bug, i.e something that will eventually work

they intentionally made like this, yes, and its a bug, but they havent fixed yet
i'd say no, because in this case they have to set up the former handler to capture unhandled exceptions in flutter web, as mentioned in this comment https://github.com/PostHog/posthog-flutter/pull/214#discussion_r2459566959 (runZonedGuarded) otherwise flutter web wont capture unhandled errors.

example/lib/main.dart
jakelacey2012: Gotcha, that makes sense, thanks for taking the time to answer my silly question @marandaneto 🙇 

Great work on these SDKs by the way @ioannisj and @marandaneto <3 super easy to work with and excited to make the switch from Sentry!

marandaneto:
> @marandaneto I feel everything is now addressed. Would you mind having another pass on it?

CI is failing, can you check CI?

lib/src/error_tracking/posthog_error_tracking_autocapture_integration.dart
```diff
@@ -0,0 +1,234 @@
...
+      final collector = details.informationCollector?.call() ?? [];
+      final information = collector.isNotEmpty
+          ? (StringBuffer()..writeAll(collector, '\n')).toString()
+          : null;
+      final library = details.library;
+      final errorSummary = details.toStringShort();
+
+      // Build additional properties with Flutter-specific details
+      final flutterErrorDetails = <String, Object>{
+        if (context != null) 'context': 'thrown $context',
```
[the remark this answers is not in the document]
ioannisj: This was intentional since based on docs https://api.flutter.dev/flutter/foundation/FlutterErrorDetails/context.html
> should be in a form that will make sense in English when following the word "thrown", as in "thrown while obtaining the image from the network" (for the context "while obtaining the image from the network").

lib/src/posthog_config.dart
```diff
@@ -44,6 +44,9 @@ class PostHogConfig {
   /// Configuration for error tracking and exception capture
   final errorTrackingConfig = PostHogErrorTrackingConfig();
 
+  /// Configuration for error tracking and exception capture
+  final errorTrackingConfig = PostHogErrorTrackingConfig();
```
marandaneto: this looks like dupe from L45

lib/src/error_tracking/isolate_handler_io.dart
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    _isolateErrorPort = RawReceivePort(onError);
+    Isolate.current.addErrorListener(_isolateErrorPort!.sendPort);
+  }
+
+  /// Remove error listener and clean up
+  void removeErrorListener() {
+    if (_isolateErrorPort != null) {
+      _isolateErrorPort!.close();
```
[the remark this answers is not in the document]
ioannisj: Assumed it was safe because we were inside a null-check block, but now I realize that since the reference is mutable it could be _theoretically_ reassigned between the check and the actual call (even though single thread execution? Still compiler requires it). With an immutable local, compiler promotes it to non-nullable type within the block. TIL :)

lib/src/posthog_config.dart
ioannisj: Yeah this was the CI error. Pbly missed it when merging main

lib/src/error_tracking/posthog_error_tracking_autocapture_integration.dart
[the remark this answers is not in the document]
ioannisj: Actually a really good point since ppl will filter by their copy-pasta string. Let's remove

marandaneto (approved):
I'd add tests for the new handlers eg isolate etc
otherwise LGTM

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 ## Next
 
+- feat: add autocapture exceptions ([#214](https://github.com/PostHog/posthog-flutter/pull/214))
+  - **Limitations**:
+    - No Flutter web support
+    - No native iOS exception capture
+    - No native C/C++ exception capture on Android (Java/Kotlin only)
```
marandaneto: we should mention proguard/minified java/kotlin no support yet here

ioannisj:
> I'd add tests for the new handlers eg isolate etc otherwise LGTM

Tried adding some but could not find a good way of unit testing these. I'll give it another shot

CHANGELOG.md
ioannisj: I thought we mention on next line though? I'm probably not understanding this well enough though

Will this line do? 
`Proguard minified stacktraces are not supported`

PR #218, opened 2025-11-07
fix: posthog-config code comments

### LINKED ISSUE #217: The default value of the config flags doesn’t match the written comments
### Version

5.9.0

### Steps to Reproduce

https://github.com/PostHog/posthog-flutter/blob/8a6f1486e5206bb5579c2b65107f9690dd10f34f/lib/src/posthog_config.dart#L187-L188
https://github.com/PostHog/posthog-flutter/blob/8a6f1486e5206bb5579c2b65107f9690dd10f34f/lib/src/posthog_config.dart#L199-L200
https://github.com/PostHog/posthog-flutter/blob/8a6f1486e5206bb5579c2b65107f9690dd10f34f/lib/src/posthog_config.dart#L211-L212

The default values for the `capturePlatformDispatcherErrors`, `captureNativeExceptions` and `captureIsolateErrors` flags are `false`. However, the comments indicate `true`, creating an inconsistency.

### Expected Result

As per comments the default values should be `true`

### Actual Result

The default value is set to `false`, which is also false 😁

ioannisj:

Closes #217 

#skip-changelog

## :pencil: Checklist

### DIFF
--- a/lib/src/posthog_config.dart
+++ b/lib/src/posthog_config.dart
@@ -164,7 +164,7 @@ class PostHogErrorTrackingConfig {
   /// **Note:**
   /// - Flutter web: Not supported
   ///
-  /// Default: true (when autocapture is enabled)
+  /// Default: false
   var captureFlutterErrors = false;
 
   /// Enable capturing of silent Flutter errors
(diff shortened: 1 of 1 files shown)


ioannisj:
Yeah just saw that which is weird since we are `latest-stable`

marandaneto:
> latest-stable

it does not mean that github already updated their latest image.
maybe we will tor force something similar on the iOS SDK eg `xcodebuild -downloadAllPlatforms` but instead of all, just what we need.
not a blocker to get this merged, but needed if we want to cut a new patch

PR #224, opened 2025-12-05
feat: add onFeatureFlags callback

ioannisj:
Based off original PR here https://github.com/PostHog/posthog-flutter/pull/183
Relates to https://github.com/PostHog/posthog-flutter/issues/81

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,6 @@
 
 - fix: TextFormField widgets were not being masked ([#227](https://github.com/PostHog/posthog-flutter/pull/227))
 
-## 5.9.1
-
-- fix: TextFormField widgets were not being masked ([#227](https://github.com/PostHog/posthog-flutter/pull/227))
-
 ## 5.9.0
 
 - feat: add autocapture exceptions ([#214](https://github.com/PostHog/posthog-flutter/pull/214))


lib/posthog_flutter_web.dart
```diff
@@ -20,8 +23,53 @@ class PosthogFlutterWeb extends PosthogFlutterPlatformInterface {
...
+        if (dartVariantsMap != null) {
+          flagVariants = dartVariantsMap
+              .map((key, value) => MapEntry(key.toString(), value));
+        }
+
+        // When posthog-js onFeatureFlags fires, it implies successful loading.
+        dartCallback(flags, flagVariants, errorsLoading: false);
+      }.toJS;
+
+      posthog!.onFeatureFlags(jsCallback);
```
marandaneto: do not use `!`, its a mutable var

lib/posthog_flutter_web.dart
```diff
@@ -20,8 +23,53 @@ class PosthogFlutterWeb extends PosthogFlutterPlatformInterface {
     );
     final PosthogFlutterWeb instance = PosthogFlutterWeb();
     channel.setMethodCallHandler(instance.handleMethodCall);
+    PosthogFlutterPlatformInterface.instance = instance;
```
marandaneto: should we clear this instance somewhere?

marandaneto: this exists before i joined, i think its due to being created using a very older version of flutter, and now our min flutter version is much higher so likely not needed anymore (better to test before removing tho).
you dont need a method channel for the web class anymore.
either refactor and test this now or create an issue?

lib/posthog_flutter_web.dart
```diff
@@ -20,8 +23,222 @@ class PosthogFlutterWeb extends PosthogFlutterPlatformInterface {
...
+    if (call.method == 'setup') {
+      // This case should ideally not be hit if Posthog().setup directly calls the overridden setup.
+      // However, to be safe, we can log or ignore.
+      // For now, let's assume direct call to overridden setup handles it.
+      return null;
+    }
+    return handleWebMethodCall(call);
+  }
+
+  @override
```
marandaneto: since this class is overriding all methods now, maybe we should address what is mentioned here https://github.com/PostHog/posthog-flutter/pull/224/files#r2601784001 and kill the web handler class altogether

lib/src/posthog_flutter_platform_interface.dart
```diff
@@ -3,6 +3,14 @@ import 'package:plugin_platform_interface/plugin_platform_interface.dart';
 import 'posthog_config.dart';
 import 'posthog_flutter_io.dart';
 
+/// Defines the callback signature for when feature flags are loaded.
+/// [flags] is a list of active feature flag keys.
+/// [flagVariants] is a map of feature flag keys to their variant values (String or bool).
+/// [errorsLoading] is true if there was an error loading flags, otherwise false or null.
+typedef OnFeatureFlagsCallback = void Function(
+    List<String> flags, Map<String, dynamic> flagVariants,
+    {bool? errorsLoading});
```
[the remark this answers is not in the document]
ioannisj: https://github.com/PostHog/posthog-flutter/issues/231

lib/posthog_flutter_web.dart
```diff
@@ -20,8 +23,222 @@ class PosthogFlutterWeb extends PosthogFlutterPlatformInterface {
...
+    if (config.onFeatureFlags != null && ph != null) {
+      final dartCallback = config.onFeatureFlags!;
+
+      // JS SDK calls with: (flags: string[], variants: Record, context?: {errorsLoading})
+      // Use optional positional param [jsContext] to handle when JS omits the 3rd argument
+      final jsCallback =
+          (JSArray jsFlags, JSObject jsFlagVariants, [JSObject? jsContext]) {
+        final List<String> flags = jsFlags.toDart.whereType<String>().toList();
+
+        Map<String, Object?> flagVariants = {};
```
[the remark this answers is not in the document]
ioannisj: Indeed, tried to clean up. However, in this case `Map<String, Object?> flagVariants = {};` was reader better to me than `var flagVariants = <String, Object?>{};`

lib/posthog_flutter_web.dart
ioannisj: Ended up dropping this because I noticed that `NativeCommunicator` is using method channel calls directly (which were error logging on web). Even though it was a simple fix I decided that it needed attention on its own cause we may have other places calling method channels directly. Another reason was to limit the scope of this PR.  
Opened an issue [here](https://github.com/PostHog/posthog-flutter/issues/232) instead

marandaneto: @ioannisj i am a bit confused here.
if we you ended dropping this, none of the changes were reverted, i am not sure if this is correct or not

lib/posthog_flutter_web.dart
```diff
@@ -13,15 +16,213 @@ class PosthogFlutterWeb extends PosthogFlutterPlatformInterface {
...
+  }
+
+  @override
+  Future<void> enable() async {
+    return handleWebMethodCall(const MethodCall('enable'));
+  }
+
+  @override
+  Future<bool> isOptOut() async {
+    final result = await handleWebMethodCall(const MethodCall('isOptOut'));
```
marandaneto: same as https://github.com/PostHog/posthog-flutter/pull/224/files#r2609882109

lib/posthog_flutter_web.dart
ioannisj: Apologies, I mean dropping the refactor to remove method channels. I don't think I pushed anything here to revert

lib/posthog_flutter_web.dart
```diff
@@ -13,15 +16,213 @@ class PosthogFlutterWeb extends PosthogFlutterPlatformInterface {
...
+  }
+
+  @override
+  Future<void> close() async {
+    return handleWebMethodCall(const MethodCall('close'));
+  }
+
+  @override
+  Future<String?> getSessionId() async {
+    final result = await handleWebMethodCall(const MethodCall('getSessionId'));
```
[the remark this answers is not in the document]
ioannisj: I think the issue here is that handleWebMethodCall returns Future<dynamic> and we need to cast the output to String. We could do `.then()` and map the return value but that would be essentially the same right?

lib/src/surveys/models/posthog_display_survey.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: avoid_dynamic_calls
```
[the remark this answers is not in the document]
ioannisj: No, no luck with `dart fix --apply`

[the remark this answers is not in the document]
marandaneto: ok lets keep this as is to move on for now

marandaneto (approved):

PR #248, opened 2026-01-08
feat: add groups parameter to capture() for event-level group association

omarbadran:
## 💡 Motivation and Context

The native iOS/Android SDKs support a `groups` parameter on `capture()` for event-level group association (see [group analytics docs](https://posthog.com/docs/product-analytics/group-analytics#link-events-to-groups)). This is already exposed in Python, Go, Node.js, and the native mobile SDKs, but was missing from the Flutter SDK.

**Use case:** When you have session-level groups (e.g. `company`) set via `group()`, but need to associate specific events with a different group (e.g. `project`) that shouldn't persist to other events.

```dart
// Session-level (sticky)
await Posthog().group(groupType: 'company', groupKey: 'company_123');

// Event-level (non-sticky) 
await Posthog().capture(
  eventName: 'file_uploaded',
  groups: {'project': 'project_456'},
);
```

## 💚 How did you test it?

- **Unit tests:** Added 3 tests in posthog_test.dart
- **Manual testing:** iOS simulator and Web (Chrome), verified `$groups` property appears correctly in PostHog events
- **Example app:** Added demo button

## 📝 Checklist

**Note:** Docs update for posthog.com may be needed after merge (adding Flutter to the "Link events to groups" section, similar to iOS/Android examples).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,5 @@
 name: CI
+
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


lib/posthog_flutter_web.dart
```diff
@@ -89,10 +89,31 @@ class PosthogFlutterWeb extends PosthogFlutterPlatformInterface {
   Future<void> capture({
     required String eventName,
     Map<String, Object>? properties,
+    Map<String, Object>? groups,
```
marandaneto: on Android and iOS: `groups: Map<String, String>`, so lets make it the same instead of `String, Object`, double check the rest of the PR

lib/src/posthog.dart
```diff
@@ -83,18 +83,34 @@ class Posthog {
...
+    if (currentScreen != null) {
+      final props = propertiesCopy ?? <String, Object>{};
+      if (!props.containsKey('\$screen_name')) {
+        props['\$screen_name'] = currentScreen;
+      }
+      return _posthog.capture(
+        eventName: eventName,
+        properties: props,
+        groups: groups,
+      );
```
marandaneto: not needed since it will return below anyway

marandaneto:
> > Thanks @omarbadran I left a few comments, once you follow up, we can merge and release it, thanks
> 
> Will do today once I'm done with some tasks.

@omarbadran any chances to wrap up this work so we can get this in?

marandaneto:
closing due to inactivity

PR #255, opened 2026-01-20
feat: add beforeSend hook

### LINKED ISSUE #155: before_send support

ioannisj:

Closes #155

Docs: https://github.com/PostHog/posthog.com/pull/14482

Adds a `beforeSend` callback to `PostHogConfig` that lets you drop or modify events before they're sent to PostHog.
- Added to capture(), screen() and captureException() 
- For $screen you can modify the screen name via the $screen_name property

Initially tried implementing this on the native side with method channels calling back to Dart in an attempt to also capture native initiated events, but hit timing issues since method channels are async but the beforeSend hook is called synchronously from native side. Also explored FFI but it added way too much complexity for the benefit. 

For now, decided to keep it dart only so that it stays fast, synchronous and reliable. Should be good for the majority of the use cases. 

Limitations in changelog

- Manual testing + modified sample project

## :pencil: Checklist

### DIFF
--- a/.github/workflows/ci-security.yaml
+++ b/.github/workflows/ci-security.yaml
@@ -27,3 +27,4 @@ jobs:
                   allowlist: |
                       actions/
                       PostHog/
+                      github/
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,98 @@
+name: "CodeQL Advanced"
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ${{ matrix.runner }}
+    permissions:
+      security-events: write
+      packages: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - language: java-kotlin
+            build-mode: manual
+            runner: ubuntu-latest
+          - language: swift
+            build-mode: manual
+            runner: macos-latest
+          - language: actions
+            build-mode: none
+            runner: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Setup Flutter (Swift)
+        if: matrix.language == 'swift'
+        uses: subosito/flutter-action@fd55f4c5af5b953cc57a2be44cb082c8f6635e8e # v2.21.0
+        with:
+          channel: 'stable'
+
+      - name: Setup Xcode (Swift)
+        if: matrix.language == 'swift'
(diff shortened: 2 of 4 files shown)


lib/src/posthog_flutter_io.dart
```diff
@@ -180,12 +207,21 @@ class PosthogFlutterIO extends PosthogFlutterPlatformInterface {
...
+      printIfDebug('[PostHog] Event dropped by beforeSend: $eventName');
+      return;
+    }
+
     try {
-      final normalizedProperties =
-          properties != null ? PropertyNormalizer.normalize(properties) : null;
+      final normalizedProperties = processedEvent.properties != null
+          ? PropertyNormalizer.normalize(
+              processedEvent.properties!.cast<String, Object>())
```
marandaneto: never use `!.`, fallback to empty array or check nullability

lib/src/posthog_config.dart
```diff
@@ -52,11 +58,47 @@ class PostHogConfig {
...
+  /// - This callback does NOT intercept native-initiated events such as:
+  ///   - Session replay events (`$snapshot`)
+  ///   - Application lifecycle events (`Application Opened`, etc.)
+  /// - Only user-provided properties are available; system properties
+  ///   (like `$device_type`, `$session_id`) are added by the native SDK at a later stage.
+  ///
+  /// **Note:**
+  /// - This callback runs synchronously on the Dart side
+  /// - Exceptions in the callback will cause the event to be sent unchanged
+  BeforeSendCallback? beforeSend;
```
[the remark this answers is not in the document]
ioannisj: Went with `List<BeforeSendCallback> beforeSend = [];` here: 
- simple API, single callback is just [fn], multiple is [fn1, fn2]
- this is usually set once on setup so I stayed away from mutation APIs like setCallback(), remoteCallback() etc

lib/src/posthog_event.dart
```diff
@@ -0,0 +1,35 @@
...
+
+  /// User properties to set on the user profile ($set).
+  ///
+  /// These properties will be merged with any existing user properties.
+  Map<String, Object?>? userProperties;
+
+  /// User properties to set only once on the user profile ($set_once).
+  ///
+  /// These properties will only be set if they don't already exist on the user profile.
+  Map<String, Object?>? userPropertiesSetOnce;
```
marandaneto: unless i am mistaken the value should be non nullable `Map<String, Object>` and not Object?

Issue #241, opened 2026-01-05
Async Feature Flags


gokberkbar:
### Problem Statement

Hi PostHog team,
I'm using the Flutter SDK and noticed that all feature flag functions (isFeatureEnabled(), getFeatureFlag(), getFeatureFlagPayload()) return Future — requiring async/await.
This is problematic in Flutter because widget build() methods are synchronous and cannot await async calls. 
Looking at your other SDKs (Web, Node, ReactNative etc.), feature flag checks are synchronous. It would be great if the Flutter SDK could offer the same.

We are fetching and caching all the feature flags in the app launch as a workaround but its unsustainable and iOS SDK already caches it. We end up with double cache.

Thanks!

### Solution Brainstorm

I believe this happens because the Flutter SDK wraps the native iOS/Android SDKs via platform channels, which are async by nature. A pure Flutter/Dart implementation would solve this.

jamesblasco:
Could we at least have a way to get all the feature flags in one async request so we can use them sync ourselves in our apps?

alimorgaan:
any updates or workarounds ?

Issue #231, opened 2025-12-10
Add full onFeatureFlags callback with flags, variants, and error state


ioannisj:
### Description

Mobile currently lacks parameters in the ‎`onFeatureFlags` callback, causing inconsistency with Web. We use a void callback that only notifies when flags are loaded, but the intention is to pass evaluated flags, variants, and any error state.

Context: https://github.com/PostHog/posthog-flutter/pull/224#discussion_r2601793593

- iOS (https://github.com/PostHog/posthog-ios/issues/647)
- Android (https://github.com/PostHog/posthog-android/issues/563)
- Flutter (this issue)
- RN

PR #328, opened 2026-03-06
feat: synchronous feature flag reads via Dart-side fetch

brunovsiqueira:
## Problem

The PostHog Flutter SDK wraps native iOS/Android SDKs via method channels, making **all feature flag reads async** (`Future<bool>`, `Future<Object?>`). This forces Flutter developers into `FutureBuilder`/`setState` patterns just to read a flag value:

```dart
// Current — awkward for UI code
FutureBuilder<bool>(
  future: Posthog().isFeatureEnabled('new-onboarding'),
  builder: (context, snapshot) {
    if (snapshot.data == true) { ... }
  },
)
```

There are other workarounds, like creating your own cache on top of the SDK when querying flags. However, these are all workarounds.

Most other PostHog SDKs (web, node, python, Android, iOS) provide **synchronous** flag reads. Flutter is the outlier. This is tracked across multiple issues: #241 (sync API), #231 (callback with data), #51 (pure Dart SDK).

## Solution

Fetch feature flags directly from PostHog's `/flags/?v=2&config=true` API on the Dart side using `package:http`, cache results in memory, and expose synchronous reads:

```dart
// New — use directly in build()
if (Posthog().isFeatureEnabledSync('new-onboarding')) {
  // show new flow
}
```

### New API

```dart
// Sync reads (Dart cache) — safe in build()
Posthog().isFeatureEnabledSync('my-flag')       // bool
Posthog().getFeatureFlagSync('my-flag')          // Object? (bool or variant string)
Posthog().getFeatureFlagResultSync('my-flag')    // PostHogFeatureFlagResult?
Posthog().getAllFeatureFlagResults()              // Map<String, PostHogFeatureFlagResult>
Posthog().isFeatureFlagsLoaded                   // bool

// New callback with flag data
config.onFeatureFlagsLoaded = (flags) { setState(() {}); };

// Existing API — completely unchanged
await Posthog().isFeatureEnabled('my-flag')      // still calls native
config.onFeatureFlags = () { ... };              // still works
```

## Architecture: Additive (not replacement)

The Dart cache runs **alongside** the native SDK — it does not replace it:

| | Async methods (`isFeatureEnabled`, etc.) | Sync methods (`isFeatureEnabledSync`, etc.) |
|---|---|---|
| **Source** | Native iOS/Android SDK (unchanged) | Dart-side HTTP cache (new) |
| **Populated by** | Native SDK lifecycle | Dart HTTP call to `/flags/?v=2` |
| **If Dart HTTP fails** | Still works from native | Returns `false`/`null` |

### Why additive instead of replacing native reads?

1. **Defensive** — if the Dart HTTP layer has a bug, existing integrations still work via native
2. **Observable** — no hidden behavioral change for existing users
3. **Reversible** — easy to switch later once the Dart path proves stable

The tradeoff is one extra HTTP call on startup (both native and Dart fetch flags independently). Open to feedback on whether we should instead replace the async methods to read from the Dart cache (single source of truth, fewer HTTP calls).

## Key Behaviors (following Android SDK patterns)

- **Request coalescing**: concurrent `reloadFeatureFlags()` calls are queued, not duplicated
- **Error merging**: when `errorsWhileComputingFlags=true`, only non-failed flags are merged into cache (preserving cached values for failed flags)
- **Quota limiting**: skips cache update when `feature_flags` is quota limited
- **Zero breaking changes**: all existing APIs and signatures preserved

## Test plan

- [x] 33 new unit tests (parsing, sync reads, HTTP, coalescing, error merging, quota limiting, callbacks)
- [x] All 126 tests pass (93 existing + 33 new)
- [x] `dart analyze` clean (only pre-existing warning)
- [ ] Manual test on iOS/Android example app
- [ ] Verify flags load on startup and sync reads work in `build()`
- [ ] Verify `identify()` triggers flag reload via native callback

### DIFF
--- a/posthog_flutter/lib/src/posthog_flutter_io.dart
+++ b/posthog_flutter/lib/src/posthog_flutter_io.dart
@@ -10,6 +10,7 @@ import 'package:posthog_flutter/src/util/logging.dart';
 import 'surveys/models/posthog_display_survey.dart' as models;
 import 'surveys/models/survey_callbacks.dart';
 import 'error_tracking/dart_exception_processor.dart';
+import 'feature_flags/posthog_feature_flags.dart';
 import 'utils/capture_utils.dart';
 import 'utils/property_normalizer.dart';
 
@@ -29,6 +30,13 @@ class PosthogFlutterIO extends PosthogFlutterPlatformInterface {
   final _methodChannel = const MethodChannel('posthog_flutter');
 
   OnFeatureFlagsCallback? _onFeatureFlagsCallback;
+  OnFeatureFlagsLoadedCallback? _onFeatureFlagsLoadedCallback;
+
+  /// Dart-side feature flags manager.
+  PostHogFeatureFlags? _featureFlags;
+
+  /// Exposes the feature flags manager for sync reads from [Posthog].
+  PostHogFeatureFlags? get featureFlags => _featureFlags;
 
   /// Stored configuration for accessing inAppIncludes and other settings
   PostHogConfig? _config;
@@ -89,7 +97,10 @@ class PosthogFlutterIO extends PosthogFlutterPlatformInterface {
         await cleanupSurveys();
         return null;
       case 'onFeatureFlagsCallback':
+        // Fire legacy callback immediately (backward compat)
         _onFeatureFlagsCallback?.call();
+        // Also reload Dart-side flags; onFeatureFlagsLoaded fires when done
+        _reloadDartFlags();
         break;
       default:
         printIfDebug(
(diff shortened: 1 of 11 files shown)


posthog_flutter/lib/src/posthog_flutter_io.dart
```diff
@@ -668,8 +698,27 @@ class PosthogFlutterIO extends PosthogFlutterPlatformInterface {
...
+  /// Reloads feature flags from the Dart side by fetching the current
+  /// distinct ID from native and making a direct API call.
+  Future<void> _reloadDartFlags() async {
+    final featureFlags = _featureFlags;
+    if (featureFlags == null) return;
+
+    try {
+      final distinctId = await getDistinctId();
+      if (distinctId.isNotEmpty) {
+        await featureFlags.loadFeatureFlags(distinctId);
```
dustinbyrne: We're missing the anonymous ID and additional evaluation properties here. This would cause some flags to evaluate differently in dart than in native.

dustinbyrne (commented):
Hey @brunovsiqueira, thanks for the pull request! 👏 

Just jotting down some initial thoughts here. From my perspective, I'd like to understand if it's necessary to introduce an additional feature flags client in Dart, or if we can cleanly get cached flag values into Dart from the native side such that we can serve them synchronously from a thin cache without requiring an additional client/request.

Open to hearing thoughts here.

OrcunJoyoLabs:
Hey folks - curious what was the outcome of this PR @brunovsiqueira did you find a workaround for yourself?

PR #347, opened 2026-03-27
feat: enable captureApplicationLifecycleEvents by default and align Android config key

marandaneto:

`captureApplicationLifecycleEvents` was disabled by default, requiring explicit opt-in. This changes the default to `true` so lifecycle events are captured out of the box.

Additionally, the Android manifest config key is aligned with iOS:
- **New key**: `com.posthog.posthog.CAPTURE_APPLICATION_LIFECYCLE_EVENTS`
- **Legacy key**: `com.posthog.posthog.TRACK_APPLICATION_LIFECYCLE_EVENTS` (still supported for backward compatibility)

The new key is checked first; if not present, the legacy key is used. If neither is set, defaults to `true`.

The changeset includes instructions on how to disable lifecycle events for users who don't want them.

Manual verification of manifest key resolution logic and default values on Android and iOS.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/warm-lions-glow.md
+++ b/.changeset/warm-lions-glow.md
@@ -2,4 +2,10 @@
 'posthog_flutter': minor
 ---
 
-Enable captureApplicationLifecycleEvents by default and align Android config key name
+Enable captureApplicationLifecycleEvents by default and align Android config key name.
+
+Application lifecycle events (`Application Opened`, `Application Backgrounded`, etc.) are now captured by default. If you don't want these events, you can disable them:
+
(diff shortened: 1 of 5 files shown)


marandaneto:
android and ios are enabled by default, so i dont see a reason why not to enable by default on ios as well

ioannisj (approved):

taxonli-yuanxin:
> android and ios are enabled by default, so i dont see a reason why not to enable by default on flutter as well

Does these events count/get billed?

marandaneto:
> > android and ios are enabled by default, so i dont see a reason why not to enable by default on flutter as well
> 
> Does these events count/get billed?

yes!

PR #220, opened 2025-11-23
Add Dio interceptor for capturing network events with Posthog

TabooSun:
## :bulb: Motivation and Context

Close #169

Didn't manage to test the code because when I check the Posthog dashboard, none of the events are recorded in the `Activity` pane, not sure if it is filtered, might need help here.

## :pencil: Checklist

- [x] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [ ] I updated the docs if needed.
- [ ] No breaking change or entry added to the changelog.

PR #358, opened 2026-04-08
feat: Add captureNetworkEvent() for HTTP telemetry in session replay

### LINKED ISSUE #169: Session replay Network performance recording
### Description

https://posthog.com/docs/session-replay/network-recording
Instrument packages such as `http` and `dio`

gabuldev:
## Summary

Resolves #169

Adds a public `Posthog().captureNetworkEvent()` method that records HTTP request metadata as rrweb custom events in the session replay timeline — enabling the **network waterfall** view for Flutter mobile apps.

### Problem

The PostHog iOS SDK uses URLSession swizzling and Android uses OkHttpInterceptor to capture network requests. Flutter/Dart uses its own HTTP stack (`http`, `dio`, `chopper`, etc.), so these native hooks never fire. The Flutter plugin even explicitly sets `captureNetworkTelemetry = false` on iOS.

### Solution

A **100% Dart** implementation with **zero new dependencies**. Instead of trying to monkey-patch HTTP clients, we expose a simple public API that developers call from their own HTTP wrapper:

```dart
// 1. Enable in config
final config = PostHogConfig('API_KEY');
config.sessionReplay = true;
config.sessionReplayConfig.captureNetworkTelemetry = true;
await Posthog().setup(config);

// 2. Call from any HTTP client
await Posthog().captureNetworkEvent(
  url: 'https://api.example.com/users',
  method: 'GET',
  statusCode: 200,
  startTimestampMs: start,
  endTimestampMs: end,
  transferSize: responseSize,
);
```

This works with **any** HTTP client — `http`, `dio`, `chopper`, `retrofit`, custom clients — without adding dependencies to the SDK.

### `posthog_config.dart`
- Added `captureNetworkTelemetry` (bool, default `false`) to `PostHogSessionReplayConfig`
- Serialized in `toMap()`

### `posthog.dart`
- Added `captureNetworkEvent()` to the `Posthog` singleton
- Parameters: `url`, `method`, `statusCode`, `startTimestampMs`, `endTimestampMs`, `transferSize`
- Guards: config flag disabled → no-op, session replay inactive → no-op, `kIsWeb` → no-op, PostHog host URL → filtered (prevents infinite capture loops)
- Formats the standard rrweb type 5 `performanceSpan` event

### `replay/native_communicator.dart`
- Added `sendNetworkEvent()` — sends `$snapshot` events directly via MethodChannel
- Bypasses the Dart `beforeSend` pipeline (consistent with how screenshots are sent)
- The native SDK automatically attaches `$session_id`

### `test/network/capture_network_event_test.dart` (new)
- 9 tests covering: correct rrweb event format, all guard paths (config disabled, session inactive, PostHog URL filtering, null config/SDK not initialized), failed requests (status 0), default transferSize

## Design decisions

| Decision | Rationale |
|----------|-----------|
| Client-agnostic API | Developers wrap their own HTTP client — no `dio`/`http` dependency in the SDK |
| Only metadata captured | URL, method, status, timing, size — no bodies/headers (privacy by design) |
| Opt-in via config flag | `captureNetworkTelemetry = false` by default — no behavior change for existing users |
| `$snapshot` via MethodChannel | Same path as screenshots — native SDK attaches session ID automatically |
| No web support | `kIsWeb` guard — the JS SDK already handles network capture on web |
| No batching | One event per request — keeps implementation simple for v1 |

## Integration examples

<details>
<summary>http package</summary>

```dart
class PostHogHttpClient extends http.BaseClient {
  final http.Client _inner;
  PostHogHttpClient(this._inner);

  @override
  Future<http.StreamedResponse> send(http.BaseRequest request) async {
    final start = DateTime.now().millisecondsSinceEpoch;
    final response = await _inner.send(request);
    final end = DateTime.now().millisecondsSinceEpoch;

    Posthog().captureNetworkEvent(
      url: request.url.toString(),
      method: request.method,
      statusCode: response.statusCode,
      startTimestampMs: start,
      endTimestampMs: end,
      transferSize: response.contentLength ?? 0,
    );
    return response;
  }
}
```
</details>

<details>
<summary>Dio interceptor</summary>

```dart
class PostHogDioInterceptor extends Interceptor {
  @override
  void onRequest(RequestOptions options, RequestInterceptorHandler handler) {
    options.extra['posthog_start'] = DateTime.now().millisecondsSinceEpoch;
    handler.next(options);
  }

  @override
  void onResponse(Response response, ResponseInterceptorHandler handler) {
    final start = response.requestOptions.extra['posthog_start'] as int;
    Posthog().captureNetworkEvent(
      url: response.requestOptions.uri.toString(),
      method: response.requestOptions.method,
      statusCode: response.statusCode ?? 0,
      startTimestampMs: start,
      endTimestampMs: DateTime.now().millisecondsSinceEpoch,
      transferSize: response.data?.toString().length ?? 0,
    );
    handler.next(response);
  }

  @override
  void onError(DioException err, ErrorInterceptorHandler handler) {
    final start = err.requestOptions.extra['posthog_start'] as int;
    Posthog().captureNetworkEvent(
      url: err.requestOptions.uri.toString(),
      method: err.requestOptions.method,
      statusCode: err.response?.statusCode ?? 0,
      startTimestampMs: start,
      endTimestampMs: DateTime.now().millisecondsSinceEpoch,
      transferSize: 0,
    );
    handler.next(err);
  }
}
```
</details>

## What this does NOT include (v1 scope)

- No native changes (Swift/Kotlin)
- No new dependencies
- No body/header capture (only metadata)
- No batching (1 event per request)
- No web support (guard returns early)

## Test plan

- [x] All 107 tests pass (`flutter test` — 98 existing + 9 new)
- [x] Network event format matches rrweb type 5 `performanceSpan` spec
- [x] All guards verified (config off, session inactive, web, PostHog URL)
- [ ] Manual test with example app + session replay on iOS/Android
- [ ] Verify network waterfall appears in PostHog dashboard

### DIFF
--- a/posthog_flutter/lib/src/posthog.dart
+++ b/posthog_flutter/lib/src/posthog.dart
@@ -9,6 +9,7 @@ import 'posthog_config.dart';
 import 'posthog_flutter_platform_interface.dart';
 import 'posthog_internal_events.dart';
 import 'posthog_observer.dart';
+import 'replay/native_communicator.dart';
 
 class Posthog {
   static PosthogFlutterPlatformInterface get _posthog =>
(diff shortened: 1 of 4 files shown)


posthog_flutter/lib/src/posthog.dart
```diff
@@ -347,6 +348,102 @@ class Posthog {
...
+  ///       statusCode: response.statusCode,
+  ///       startTimestampMs: start,
+  ///       endTimestampMs: end,
+  ///       transferSize: response.contentLength ?? 0,
+  ///     );
+  ///     return response;
+  ///   }
+  /// }
+  /// ```
+  Future<void> captureNetworkEvent({
```
marandaneto, line 394: this is not an event so it should not be part of the public API, not on this class at least, or named like this
ideally we'd have something like this https://github.com/PostHog/posthog-flutter/pull/220 so users dont have to do anything, i just didnt have time to do that just yet

maybe a better approach for now would be a class that transforms the metadata into an event and then you just call Posthog.capture like:

val networkSnapshot = PostHogNetworkSnapshotEvent(...)
PostHog.capture('$snapshot', networkSnapshot) 
or
PostHog.capture(networkSnapshot.event, networkSnapshot) 
or just networkSnapshot.capture()

my point here is more about reusing `PostHog.capture` if possible

marandaneto: if users intentionally call this method, we dont need to gate against captureNetworkTelemetry

marandaneto: captureNetworkTelemetry would be more if we offer some sort of auto-capture network snapshots

marandaneto:
@gabuldev would you be able to address the comments so we can get this merged?

PR #364, opened 2026-04-15
fix: reduce session replay frame callback overhead

marandaneto:

`ChangeDetector` previously registered an `addPostFrameCallback` after every rendered frame, even though session replay only attempted a capture at the configured throttle interval. That meant the app was still doing scheduler work roughly every frame just to reject most callbacks in `PostHogWidget`.

This change moves cadence into `ChangeDetector` itself so it only schedules post-frame callbacks at the configured interval, while `PostHogWidget` keeps only the in-flight capture guard.

- `flutter analyze posthog_flutter`
- Manual review of the `ChangeDetector` polling flow and `PostHogWidget` capture guard

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/posthog_flutter/lib/src/replay/change_detector.dart
+++ b/posthog_flutter/lib/src/replay/change_detector.dart
@@ -1,10 +1,12 @@
+import 'dart:async';
+
 import 'package:flutter/widgets.dart';
 
 /// A class that detects changes in the UI and executes a callback when changes occur.
 ///
-/// The `ChangeDetector` continuously monitors the Flutter widget tree by scheduling
-/// a callback after each frame is rendered. This is useful when you need to perform
-/// an action whenever the UI updates.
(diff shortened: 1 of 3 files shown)


posthog_flutter/lib/src/replay/change_detector.dart
```diff
@@ -16,47 +18,54 @@ import 'package:flutter/widgets.dart';
...
-      });
+    if (_isRunning) {
+      return;
     }
+
+    _isRunning = true;
+    _scheduleFrameCallback();
+    _timer = Timer.periodic(interval, (_) {
+      _scheduleFrameCallback();
+    });
```
dustinbyrne, line 47: Is it possible for this timer to fire faster than the frame rate?

Assuming `WidgetsBinding.instance.addPostFrameCallback` allows callbacks to queue up

dustinbyrne (approved):

PR #366, opened 2026-04-15
fix(web): accept Map args in startSessionRecording handler

### LINKED ISSUE #365: [Bug] Flutter Web: startSessionRecording throws TypeError — handler expects bool but platform sends Map
## Bug description

On Flutter Web, calling `Posthog().startSessionRecording()` throws a `TypeError` and crashes the frame:

```
TypeError: Instance of 'LinkedHashMap<String, bool>': type 'LinkedHashMap<String, bool>' is not a subtype of type 'bool?'
```

The web method-call handler casts the method arguments as `bool?`, but the web platform interface passes them as a `Map<String, dynamic>`.

## Affected versions

Reproduced on `5.15.2`, `5.23.0`, `5.23.1`, and current `main` (commit `27ffbc1`). Introduced by #256 (`feat: add manual start and stop recording`).

## Stack trace (dart2js release build, captured in production via Sentry)

```
_TypeError._throwUnexpectedSubtype        (runtime.dart)
_TypeError._throwSubtype                  (runtime.dart)
rti._asBoolQ                              (rti.dart:1639)
handleWebMethodCall                       (posthog_flutter_web_handler.dart:295)
PosthogFlutterWeb.startSessionRecording   (posthog_flutter_web.dart:333)
Posthog.startSessionRecording             (posthog.dart:375)
```

## Root cause

`posthog_flutter/lib/posthog_flutter_web.dart` (line ~333):

```dart
@override
Future<void> startSessionRecording({bool resumeCurrent = true}) async {
  return handleWebMethodCall(
    MethodCall('startSessionRecording', {'resumeCurrent': resumeCurrent}),
  );
}
```

sends a `Map`, but `posthog_flutter/lib/src/posthog_flutter_web_handler.dart:295` reads:

```dart
case 'startSessionRecording':
  final resumeCurrent = args as bool? ?? true;  // args is a Map, not a bool
```

Every other `case` in the handler reads `args['someKey']`; this one was missed. In debug builds this is just a runtime type error; in dart2js release builds the minified `LinkedHashMap` message (`minified:nh<String, bool>`) makes the origin harder to spot but the root cause is the same.

## Repro

1. Initialize `posthog-js` on the page.
2. From Flutter Web, call `Posthog().startSessionRecording();`.
3. The call throws (visible in the console and propagated to Zone error handlers / Sentry).

## Proposed fix

Read `resumeCurrent` from the Map (matching every other `case` in the handler) and keep the legacy `bool` path as a fallback:

```dart
case 'startSessionRecording':
  final bool resumeCurrent;
  if (args is Map) {
    resumeCurrent = args['resumeCurrent'] as bool? ?? true;
  } else {
    resumeCurrent = args as bool? ?? true;
  }
  ...
```

PR is on the way.

## Environment

- Platform: Flutter Web (dart2js release and debug)
- `posthog_flutter`: 5.23.1 (same bug in 5.15.2 through current `main`)
- Found via a production Sentry crash affecting real web users.

Ortes:
## Summary

- Fixes a `TypeError` on Flutter Web when calling `Posthog().startSessionRecording()`: the platform interface dispatches a `Map<String, dynamic>` (`{'resumeCurrent': bool}`) but the web handler casts the arguments as `bool?`, crashing on every call.
- Reads `resumeCurrent` from the Map (matching every other `case` in the handler) and keeps the legacy `bool` path as a fallback so older callers still work.
- Closes #365.

## Details

`posthog_flutter/lib/posthog_flutter_web.dart`:

```dart
@override
Future<void> startSessionRecording({bool resumeCurrent = true}) async {
  return handleWebMethodCall(
    MethodCall('startSessionRecording', {'resumeCurrent': resumeCurrent}),
  );
}
```

`posthog_flutter/lib/src/posthog_flutter_web_handler.dart:295` (before):

```dart
case 'startSessionRecording':
  final resumeCurrent = args as bool? ?? true;  // crashes: args is a Map
```

After:

```dart
case 'startSessionRecording':
  final bool resumeCurrent;
  if (args is Map) {
    resumeCurrent = args['resumeCurrent'] as bool? ?? true;
  } else {
    resumeCurrent = args as bool? ?? true;
  }
```

## Test plan

- [x] Verified the failing cast is reached in release (dart2js) via a production Sentry stack trace: `_asBoolQ` → `handleWebMethodCall` at `posthog_flutter_web_handler.dart:295`.
- [x] Patched handler reads `resumeCurrent` from the Map sent by `PosthogFlutterWeb.startSessionRecording` and falls back to the legacy `bool` shape.
- [ ] Reviewers to confirm there is no other caller still passing a raw `bool` that relies on the previous cast.

### DIFF
--- a/posthog_flutter/lib/src/posthog_flutter_web_handler.dart
+++ b/posthog_flutter/lib/src/posthog_flutter_web_handler.dart
@@ -292,7 +292,14 @@ Future<dynamic> handleWebMethodCall(MethodCall call) async {
     case 'isSessionReplayActive':
       return posthog?.sessionRecordingStarted() ?? false;
     case 'startSessionRecording':
-      final resumeCurrent = args as bool? ?? true;
+      // The platform interface sends a Map ({'resumeCurrent': bool}). Fall back
+      // to the legacy bool shape so older callers keep working.
+      final bool resumeCurrent;
+      if (args is Map) {
+        resumeCurrent = args['resumeCurrent'] as bool? ?? true;
+      } else {
+        resumeCurrent = args as bool? ?? true;
+      }
       if (!resumeCurrent) {
         // Reset session ID to start a new session
         posthog?.sessionManager?.resetSessionId();


marandaneto:
@ioannisj can you double check this? i wonder if that makes sense since we didnt support startSessionRecording before on the web

ioannisj (approved):
LG, yeah I think this was an oversight

marandaneto:
i cannot merge this as is because theres no signed commits

marandaneto:
thanks @Ortes 
https://github.com/PostHog/posthog-flutter/pull/368

PR #374, opened 2026-04-22
Rename apiKey to projectToken with aliases

marandaneto:

The Flutter SDK used `apiKey` for the project credential while other SDKs are moving toward more explicit naming. This change introduces `projectToken` as the preferred name in the Dart API and native auto-init configuration, while keeping the previous `apiKey` / `API_KEY` names as fallbacks for backwards compatibility.

This follow-up also improves deprecation logging for legacy `apiKey` / `API_KEY` usage and makes missing-key logs clearer by explicitly saying either the new or legacy key must be provided.

- Ran `flutter test test/posthog_test.dart test/posthog_flutter_io_test.dart` in `posthog_flutter/`
- Ran `flutter analyze lib test` in `posthog_flutter/`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
+++ b/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
@@ -59,10 +59,13 @@ class PosthogFlutterPlugin :
                 return
             }
 
-            val apiKey = bundle.getString("com.posthog.posthog.API_KEY")?.trim()
+            val projectToken =
+                (bundle.getString("com.posthog.posthog.PROJECT_TOKEN")
+                    ?: bundle.getString("com.posthog.posthog.API_KEY"))
+                    ?.trim()
 
-            if (apiKey.isNullOrEmpty()) {
-                Log.e("PostHog", "com.posthog.posthog.API_KEY is missing!")
+            if (projectToken.isNullOrEmpty()) {
+                Log.e("PostHog", "com.posthog.posthog.PROJECT_TOKEN/com.posthog.posthog.API_KEY is missing!")
                 return
             }
 
@@ -79,7 +82,8 @@ class PosthogFlutterPlugin :
             val debug = bundle.getBoolean("com.posthog.posthog.DEBUG", false)
 
             val posthogConfig = mutableMapOf<String, Any>()
-            posthogConfig["apiKey"] = apiKey
+            posthogConfig["projectToken"] = projectToken
+            posthogConfig["apiKey"] = projectToken
             posthogConfig["host"] = host
             posthogConfig["captureApplicationLifecycleEvents"] = captureApplicationLifecycleEvents
             posthogConfig["debug"] = debug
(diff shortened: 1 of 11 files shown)


posthog_flutter/lib/src/posthog_config.dart
```diff
@@ -17,7 +17,17 @@ enum PostHogDataMode { wifi, cellular, any }
...
+  /// Your PostHog project token.
+  ///
+  /// You can find it at:
+  /// https://us.posthog.com/settings/project-details#variables
+  ///
+  /// This field was formerly named [apiKey].
+  final String projectToken;
+
+  @Deprecated('Deprecated in favor of [projectToken].')
+  String get apiKey => projectToken;
```
marandaneto, line 30: this was set via ctor so all good as readonly

marandaneto:
@rafaeelaudibert will try to migrate sdks that wont break anything, this is only possible with sdks that have positional parameters and no named parameters
will at least add a comment pointing to https://us.posthog.com/settings/project-details#variables that this is now called project token and used to be called something else

turnipdabeets (approved):

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -59,10 +59,13 @@ class PosthogFlutterPlugin :
                 return
             }
 
-            val apiKey = bundle.getString("com.posthog.posthog.API_KEY")?.trim()
+            val projectToken =
+                (bundle.getString("com.posthog.posthog.PROJECT_TOKEN")
+                    ?: bundle.getString("com.posthog.posthog.API_KEY"))
+                    ?.trim()
```
rafaeelaudibert, line 65: Always fun learning new syntax :). Is this equivalent to TS's `??`?

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -59,10 +59,13 @@ class PosthogFlutterPlugin :
                 return
             }
 
-            val apiKey = bundle.getString("com.posthog.posthog.API_KEY")?.trim()
+            val projectToken =
```
rafaeelaudibert, line 62: Is it worth adding a deprecation notice for `API_KEY`? And then we can drop the notice + the old usage when/if we release a new major.

Since this is the first PR I see for this we should figure out what the standard should be. This also means we'd be able to implement this for named arguments, we just issue a warning if people use the old naming

rafaeelaudibert (approved):

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
marandaneto: yes, elvis operator

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
marandaneto: `@Deprecated('Deprecated in favor of [projectToken].')` native linters/IDE will warn at build time/usage already

its a standard that this would be removed in the next major

rafaeelaudibert: I saw that after I'd posted this one already, all good! I'd still be in favor of an actual log during runtime, but that's all right

marandaneto: https://github.com/PostHog/posthog-flutter/pull/374/changes/4a33e1404b639f7b727865883f5d8de48924eb4b
we have both now

PR #392, opened 2026-05-22
chore: migrate Android example to AGP 9

### LINKED ISSUE #390: Migrate Plugin to Built-in Kotlin
### Problem Statement

Starting with Android Gradle Plugin (AGP) 9.0, support for applying the Kotlin Gradle Plugin (KGP) has been removed. Because this plugin applies KGP, it causes a compilation error that prevents my app from building. Here is an example of the error https://github.com/flutter/flutter/issues/181383.

Flutter has temporarily added support to allow KGP while apps and plugins migrate to AGP 9.0+, but this support will be removed in a future version of Flutter.

Please migrate this plugin to use built-in Kotlin to ensure your plugin users can successfully build their apps in future versions of Flutter.

Here is the Flutter [migration guide for plugin authors](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors).

Please be respectful and mindful of the plugin repository's rules and code of conduct when reporting issues and interacting with plugin authors.

For reference, see the [Flutter Code of Conduct](https://github.com/flutter/flutter/blob/master/CODE_OF_CONDUCT.md).

### Solution Brainstorm

_No response_

marandaneto:

Migrate the Android example app to AGP 9.2.1 and Gradle 9.4.1, while aligning the example Android project structure with the latest Flutter Android template.

Fixes #390.

- `cd example/android && ./gradlew help --no-daemon --console=plain`
- `cd example/android && ./gradlew :app:compileDebugKotlin --no-daemon --console=plain`
- `cd example/android && ./gradlew :app:assembleRelease --no-daemon --console=plain`
- `cd example/android && ./gradlew build --dry-run --no-daemon --console=plain`
- `cd example && flutter build apk --release`
- Verified `posthog_flutter/android` Gradle configuration with AGP 8.11.1 using Gradle 8.14 and an injected Flutter `compileSdkVersion` property

## :pencil: Checklist

### If releasing new changes


pubspec.lock
```diff
@@ -230,5 +230,5 @@ packages:
     source: hosted
     version: "1.1.1"
 sdks:
-  dart: ">=3.11.0 <4.0.0"
-  flutter: ">=3.41.0"
+  dart: ">=3.12.0 <4.0.0"
+  flutter: ">=3.44.0"
```
marandaneto: is this for the example or also for posthog_flutter? it should be for the example only

posthog_flutter/android/build.gradle
```diff
@@ -69,3 +68,9 @@ android {
         }
     }
 }
+
+kotlin {
+    compilerOptions {
+        jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8
+    }
+}
```
marandaneto: is this compatible with AGP 8 as well?
we can import org.jetbrains.kotlin.gradle.dsl.JvmTarget so we dont need to add the full namespace here

example/android/app/build.gradle
```diff
@@ -26,36 +25,28 @@ if (flutterVersionName == null) {
...
         targetCompatibility JavaVersion.VERSION_1_8
     }
 
-    kotlinOptions {
-        jvmTarget = '1.8'
-    }
-
-    sourceSets {
-        main.java.srcDirs += 'src/main/kotlin'
-    }
```
marandaneto, line 43: we dont need this anymore?

pubspec.lock
marandaneto: Fixed in 978c897 — reverted the root workspace pubspec.lock change so the SDK requirement bump stays scoped to the example pubspec.

posthog_flutter/android/build.gradle
marandaneto: Fixed in 978c897 — imported JvmTarget and used the short reference. I also verified the posthog_flutter Android Gradle config still configures with its AGP 7.4.2/KGP 2.0.0 setup, so this remains compatible with AGP 8.x as well. The compilerOptions DSL comes from KGP 2.0.0 rather than AGP.

example/android/app/build.gradle
marandaneto: This should be safe to remove for the AGP 9 example migration. AGP/built-in Kotlin recognizes src/main/kotlin by default, so we no longer need to add it through the Java source set. Verified :app:compileDebugKotlin passes.

marandaneto:
Fixed the release Kotlin target mismatch in 79fdc59 by configuring all KotlinCompile tasks to use JVM 1.8 for both the AGP 9 example app and the plugin's AGP 7/8 path. Also added the missing example ProGuard rules file required by AGP 9 release minification. Verified :app:assembleRelease now passes.

marandaneto:
Pushed f94acba to explicitly keep built-in Kotlin enabled for the migrated example app. Flutter 3.44's migrator otherwise appends android.builtInKotlin=false, which opts the example back out of the migration. I verified release assembly with built-in Kotlin enabled, including with a stale ignored GeneratedPluginRegistrant.java present locally.

marandaneto:
Fixed the Flutter APK lookup failure in 8f7acfb. The Gradle 9 layout.buildDirectory migration was resolving to example/android/build instead of example/build; changed it to ../../build so Flutter finds build/app/outputs/flutter-apk/app-release.apk again. Verified with cd example && flutter build apk --release.

posthog_flutter/android/build.gradle
```diff
@@ -69,3 +71,9 @@ android {
         }
     }
 }
+
+tasks.withType(KotlinCompile).configureEach {
```
marandaneto, line 84: what if this is AGP 9 and we dont do apply plugin: 'kotlin-android'
will this line still work? KotlinCompile comes from org.jetbrains.kotlin.gradle.tasks, isnt that part of kotlin-android?

posthog_flutter/android/build.gradle
```diff
@@ -22,7 +25,18 @@ allprojects {
...
 apply plugin: 'com.android.library'
-apply plugin: 'kotlin-android'
+
+def agpMajor = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0] as int
+if (agpMajor < 9) {
+    apply plugin: 'kotlin-android'
+} else {
+    def builtInKotlin = providers.gradleProperty('android.builtInKotlin')
+        .map { it.toBoolean() }
+        .getOrElse(true)
```
ioannisj, line 37: I would expect this to default to false if it's not explicitly set?

ioannisj (approved):

posthog_flutter/android/build.gradle
marandaneto: Good question. For AGP 9 itself, built-in Kotlin defaults to true when the property is not set. The current Flutter template explicitly opts out with android.builtInKotlin=false, and this branch handles that by applying kotlin-android only when the property is explicitly false. I added a comment in b1b199b to make that default/opt-out behavior clear.

.changeset/quiet-ducks-build.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog_flutter': minor
+---
+
+Support Android builds with AGP 9 built-in Kotlin while preserving compatibility with AGP 8 and earlier. This release requires Android Gradle Plugin 8.0 or newer.
```
marandaneto:
```suggestion
Support Android builds with AGP 9 built-in Kotlin while preserving compatibility with AGP 8. This release requires Android Gradle Plugin 8.0 or newer.
```

PR #393, opened 2026-05-23
feat: capture platform views in session replay screenshots

### LINKED ISSUE #151: Session Recordings: Platform Views (Native Views)
### Problem Statement

it seems that session recordings do not include recording native platform views. 

We are using Superwall to test our paywall which presents a native iOS view on top of the existing Flutter view. However, in session recordings - Posthog only displays the flutter view and does not show the Paywall

### Solution Brainstorm

_No response_

marcorizza:

Fixes [#151](https://github.com/PostHog/posthog-flutter/issues/151).

Session Replay in Flutter was only capturing the Flutter render tree, so native platform views rendered on top of Flutter content were missing from recordings. In practice, this meant native overlays such as paywalls, maps, or web views could appear blank or invisible in session replay snapshots.

This PR adds an opt-in `sessionReplayConfig.capturePlatformViews` flag for session replay. When enabled, Flutter requests a screenshot from the native view hierarchy on iOS and Android and uses that image for replay snapshots, allowing native platform views to be included in recordings.

The option defaults to `false` because native platform view content cannot be automatically masked using Flutter widget masking rules.

- Added a unit test to verify that `sessionReplayConfig.capturePlatformViews` is disabled by default and correctly serialized in the replay config map.
- Manually verified on iOS and Android that native platform views are captured in session replay when `sessionReplayConfig.capturePlatformViews` is enabled.
- Confirmed the existing Flutter-only behavior is preserved when the option is left disabled.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/fiery-lamps-guess.md
+++ b/.changeset/fiery-lamps-guess.md
@@ -1,0 +1,5 @@
+---
+"posthog_flutter": minor
+---
+
+Add opt-in session replay support for native platform views on iOS and Android.
--- a/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
+++ b/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
@@ -30,6 +30,7 @@ import io.flutter.plugin.common.MethodChannel.MethodCallHandler
 import io.flutter.plugin.common.MethodChannel.Result
 import java.io.ByteArrayOutputStream
 import java.util.Date
+import java.util.concurrent.Executors
 import kotlin.math.roundToInt
 
 /** PosthogFlutterPlugin */
@@ -46,6 +47,8 @@ class PosthogFlutterPlugin :
     private lateinit var applicationContext: Context
     private var activity: Activity? = null
 
+    private val mainHandler = Handler(Looper.getMainLooper())
+    private val screenshotCompressionExecutor = Executors.newSingleThreadExecutor()
     private val snapshotSender = SnapshotSender()
 
     // The surveys delegate
@@ -448,6 +451,7 @@ class PosthogFlutterPlugin :
 
     override fun onDetachedFromEngine(binding: FlutterPlugin.FlutterPluginBinding) {
         channel.setMethodCallHandler(null)
+        screenshotCompressionExecutor.shutdown()
     }
 
     private fun handleSendFullSnapshot(
@@ -553,7 +557,7 @@ class PosthogFlutterPlugin :
                 bitmap,
                 { copyResult ->
                     if (copyResult == PixelCopy.SUCCESS) {
-                        result.success(bitmapToPng(bitmap))
+                        compressBitmapToPngAsync(bitmap, result)
                     } else {
                         bitmap.recycle()
                         captureNativeScreenshotFallback(
@@ -568,7 +572,7 @@ class PosthogFlutterPlugin :
                         )
                     }
                 },
-                Handler(Looper.getMainLooper()),
+                mainHandler,
             )
             return
         }
@@ -597,6 +601,8 @@ class PosthogFlutterPlugin :
     ) {
         val contentBitmap =
             Bitmap.createBitmap(contentView.width, contentView.height, Bitmap.Config.ARGB_8888)
+        // Software Canvas cannot read GPU-composited surfaces like SurfaceView or TextureView,
+        // so those platform views may still appear blank when we hit this fallback path.
         contentView.draw(android.graphics.Canvas(contentBitmap))
 
         val croppedBitmap =
@@ -618,7 +624,7 @@ class PosthogFlutterPlugin :
                 }
             }
 
-        result.success(bitmapToPng(outputBitmap))
+        compressBitmapToPngAsync(outputBitmap, result)
     }
 
     private fun bitmapToPng(bitmap: Bitmap): ByteArray {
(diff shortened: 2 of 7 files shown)


ioannisj:
Hey, thank you for the work here! 

I'll have a deeper look soon but at first glance my biggest concern is PII leaking from platform views when the option is enabled. 

Can PostHogMaskWidget be used for those cases, so at least developers can have control over opt-in masking of platform views? 

Also I wonder if somehow exposing [findMaskableWidgets](https://github.com/PostHog/posthog-ios/blob/4a3298f4fddda0bd0229e0f06c2f80b2985aab10/PostHog/Replay/PostHogReplayIntegration.swift#L659) could help, since this will return masking rects from native. Or maybe just grab a screenshot directly from native sdks with masking already applied (both require refactor of posthog-ios and posthog-android)

Curious to hear thoughts from @marandaneto @turnipdabeets as well

marandaneto:
i think completely replacing the flutter approach with a native approach isn't great because of masking, afaik its not possible to mask on the native side since we dont have access to the maskable widgets

imo a better approach is:

stitching recordings, the flutter and the native one

theres no automatic way to find out if a screen is Flutter or native (via platform views or pure native)
the only way is with a manual API where the user would tell -> hey here starts a platform view and/or a pure native view
so we can stitch and continue a recording from the native side

eg
```
posthog.startSessionRecording() // usually starts automatically

// user calls - pseudo code
posthog.stopSessionRecording()
posthog.startNativeSessionRecording(resume: true)
showPlatformViewUI()

closePlatformViewUI()
posthog.stopNativeSessionRecording()
posthog.startSessionRecording(resume: true)

so both recordings, from both sides (flutter and native), carry the same session id and are a single playable recording
```

turnipdabeets:
Hey @marcorizza thanks for this contribution! 

I think I agree with @marandaneto here:

> imo a better approach is:
> 
> stitching recordings, the flutter and the native one

Would you be interested in trying this out?

marcorizza:
Thanks for the feedback and for the explanation.

I agree that PII/masking is the main concern with the current screenshot-based approach, especially because platform view content cannot be masked automatically through Flutter widget rules.

I’m open to exploring the stitching approach, but I’d like to better understand how it should work for common Flutter use cases involving platform views, especially `google_maps_flutter`.

For example, many Flutter apps don’t show the map as a fully native screen. The map is usually embedded inside a Flutter-rendered screen:

```dart
Scaffold(
  appBar: AppBar(...),
  body: GoogleMap(...),
  bottomNavigationBar: BottomNavigationBar(...),
)
```

or combined with Flutter overlays:

```dart
Stack(
  children: [
    GoogleMap(...),
    SearchBar(...),
    FloatingActionButton(...),
  ],
)
```

In these cases, the screen contains both a native platform view and Flutter-rendered UI at the same time.

If we switch entirely to native replay while the map is visible, wouldn’t we lose the surrounding Flutter UI such as the app bar, bottom navigation, overlays, buttons, etc.? And with the current Flutter-only replay approach, the map itself is missing from the recording.

turnipdabeets:
> Thanks for the feedback and for the explanation.
> 
> I agree that PII/masking is the main concern with the current screenshot-based approach, especially because platform view content cannot be masked automatically through Flutter widget rules.
> 
> I’m open to exploring the stitching approach, but I’d like to better understand how it should work for common Flutter use cases involving platform views, especially `google_maps_flutter`.
> 
> For example, many Flutter apps don’t show the map as a fully native screen. The map is usually embedded inside a Flutter-rendered screen:
> 
> ```dart
> Scaffold(
>   appBar: AppBar(...),
>   body: GoogleMap(...),
>   bottomNavigationBar: BottomNavigationBar(...),
> )
> ```
> 
> or combined with Flutter overlays:
> 
> ```dart
> Stack(
>   children: [
>     GoogleMap(...),
>     SearchBar(...),
>     FloatingActionButton(...),
>   ],
> )
> ```
> 
> In these cases, the screen contains both a native platform view and Flutter-rendered UI at the same time.
> 
> If we switch entirely to native replay while the map is visible, wouldn’t we lose the surrounding Flutter UI such as the app bar, bottom navigation, overlays, buttons, etc.? And with the current Flutter-only replay approach, the map itself is missing from the recording.

Thanks @marcorizza - you make a good point. Stitching can't cover a map embedded in a Scaffold. It may be that we need both stitching and screenshot.

  @marandaneto - looking at this PR closer, it actually uses the native layer to supply the pixels and mask rects are still computed from the widget tree in Dart and painted over the screenshot in the existing pipeline. We’d still have an issue with fully native screens but wondering if that’s a separate task since it requires work in posthog-ios/posthog-android. wdyt? Also tagging @ioannisj for his opinions
  
  For PII masking in this PR we could use: 
1. Flutter widgets — the SDK already computes mask rects from the widget tree (maskAllTexts, PostHogMaskWidget, etc.) and paints them over the image. Since you routed the native screenshot through the existing pipeline, this part already works in your PR (we should just verify the rect scaling lines up with the native image's pixel scale).

2. Platform view content — nothing can see inside a webview or a map, so rather than trying to detect what's sensitive, mask platform views by default even when capturePlatformViews is enabled, with an explicit opt-out for views the developer judges safe (e.g. wrapping the map in an unmask widget, or an allow-list in config). That way enabling the flag never silently leaks webview content — the developer consciously unmasks each view.

One more iOS-specific thing: since drawHierarchy captures the whole key window, a natively-presented screen (modal/sheet/paywall) would get captured unmasked if Flutter is still producing frames behind it — no widget rects cover it. Could we skip the native screenshot when Flutter isn't the topmost content (e.g. rootViewController.presentedViewController != nil) and fall back to the Flutter-only render? Android isn't affected since native screens live in their own window

@marcorizza if this works for you and we're able to address PII then I think we can continue with your screenshot approach. Let's also see what   @marandaneto and @ioannisj have to say.

turnipdabeets:
@marcorizza + @PostHog/team-client-libraries I plan to experiment with this PR this week and I may push some commits. There's a lot here I want to keep (the platform-view detection especially). The main thing I want to try is using the Flutter render as the base and compositing native content into the platform-view regions, rather than replacing the whole frame with a native screenshot. 

I also think #151 is really two separate problems: 
1. Embedded platform views (what this PR targets)
2. Fully native / presented views (e.g. the Superwall paywall)

They likely need a different approach and probably a separate PR

turnipdabeets:
@marcorizza I ended up opening a separate PR, please check it out to make sure it works for you! https://github.com/PostHog/posthog-flutter/pull/453

PR #408, opened 2026-06-01
fix: render link-type survey questions without a URL on Android

### LINKED ISSUE #407: Surveys: link-type question with no URL silently fails to render on Android (null is not a subtype of String in deserializer)
## Summary

A survey whose question is of type `link` with **no URL** renders fine on **iOS** but **silently fails to render on Android** — the bottom sheet never appears, and no error is surfaced.

## Root cause

In `lib/src/surveys/models/posthog_display_survey.dart`, `PostHogDisplaySurvey.fromDict` parses link questions with a **non-nullable cast**:

```dart
case 'link':
  return PostHogDisplayLinkQuestion(
    id: id,
    question: question,
    link: q['link'] as String,   // <-- throws when link is null
    ...
```

When a `link` question has no URL, the native layer forwards `null` on **Android** (iOS forwards `""`). `q['link'] as String` then throws `type 'Null' is not a subtype of type 'String'` inside `fromDict` — which runs in the `showSurvey` method-channel handler *before* the bottom sheet is built — so the **entire survey is dropped**. Nothing is shown and no error reaches app code.

This is inconsistent with the rest of the codebase: the model field is `final String link;` (non-nullable), but the widget `lib/src/surveys/widgets/link_question.dart` already declares `final String? link;`, and `survey_bottom_sheet.dart` only opens the URL `if (link.isNotEmpty)`. So a link-less intro/welcome screen is clearly intended to be valid — only the deserializer rejects it.

## Steps to reproduce

1. Create a survey whose first question is type `link` with an empty/absent link URL (a common pattern for an intro/welcome screen with a "Start" button).
2. Show it on **Android** → survey never appears.
3. Show the same survey on **iOS** → renders correctly.

## Expected

Link questions without a URL render on Android just like on iOS (the button simply advances to the next question, per `survey_bottom_sheet.dart`'s `if (link.isNotEmpty)` guard).

## Suggested fix

Make the parse null-safe (and the model field nullable, matching the widget):

```dart
// posthog_display_survey.dart
link: q['link'] as String? ?? '',

// posthog_display_link_question.dart
final String? link;   // widget already uses String?
```

## Environment

- `posthog_flutter` 5.24.2
- native `posthog-android` 3.x
- Android only — iOS unaffected

turnipdabeets:

Fixes #407.

A survey whose question is type `link` with no URL renders on iOS but silently fails on Android. Android forwards `null` for a missing link (iOS forwards `""`), so the non-null cast `q['link'] as String` in `PostHogDisplaySurvey.fromDict` throws and the entire survey is dropped before the bottom sheet is built — no error reaches app code.

The fix parses a missing link as an empty string (`as String? ?? ''`), normalizing both platform payloads. This matches the existing `if (link.isNotEmpty)` guard in `survey_bottom_sheet.dart`, so a link-less intro/welcome screen simply advances to the next question.

Added `posthog_display_survey_test.dart` covering `fromDict` for a link question across both platform shapes: iOS (`""`), Android (`null`, the crash case — asserts `returnsNormally`), absent key, and a real URL passed through unchanged.

## :pencil: Checklist

### If releasing new changes


posthog_flutter/test/posthog_display_survey_test.dart
```diff
@@ -0,0 +1,54 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:posthog_flutter/src/surveys/models/posthog_display_survey.dart';
+import 'package:posthog_flutter/src/surveys/models/posthog_display_link_question.dart';
+
+void main() {
+  // Builds a minimal survey dict (as forwarded by the native method channel)
+  // whose single question is of type `link`, overriding only the `link` value.
+  Map<String, dynamic> surveyWithLink(Object? link) {
```
marandaneto: Always use Object or Object? Instead of dynamic

posthog_flutter/test/posthog_display_survey_test.dart
```diff
@@ -0,0 +1,54 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:posthog_flutter/src/surveys/models/posthog_display_survey.dart';
+import 'package:posthog_flutter/src/surveys/models/posthog_display_link_question.dart';
+
+void main() {
```
marandaneto: I'd rename this file to something more generic like survey_tests instead of a file name specific to this issue only

posthog_flutter/test/posthog_display_survey_test.dart
```diff
@@ -0,0 +1,54 @@
...
+  Map<String, dynamic> surveyWithLink(Object? link) {
+    final question = <String, dynamic>{
+      'type': 'link',
+      'question': 'Welcome',
+      'isOptional': false,
+    };
+    // Mirror the native bridges: Android forwards a raw value (possibly null),
+    // so a missing URL is represented by the key being present with `null`.
+    // Passing the sentinel below lets a caller omit the key entirely.
+    if (link != _absent) {
```
marandaneto: I dont follow the sentinel usage, that could be just a method param if it should be ser and null, not set at all or with a valid value
A global var for this flow control feels weird

posthog_flutter/test/posthog_display_survey_test.dart
turnipdabeets: updated

posthog_flutter/test/posthog_display_survey_test.dart
turnipdabeets: renamed

posthog_flutter/test/posthog_display_survey_test.dart
turnipdabeets: dropped it for a named param

posthog_flutter/lib/src/surveys/models/posthog_display_survey.dart
```diff
@@ -35,7 +35,7 @@ class PostHogDisplaySurvey {
           return PostHogDisplayLinkQuestion(
             id: id,
             question: question,
-            link: q['link'] as String,
+            link: q['link'] as String? ?? '',
```
marandaneto, line 38: This fixes the casting, i am just not sure if now this makes a broken UI or not
the question is of the type `PostHogDisplayLinkQuestion`, but theres no `link`, so how does it render in the UI? is the question actionable? Or is it rendering a button that is not clickable?

marandaneto (approved):
approving but worth a follow up here https://github.com/PostHog/posthog-flutter/pull/408/changes#r3336094412

posthog_flutter/lib/src/surveys/models/posthog_display_survey.dart
turnipdabeets: UI isn't broken, this now matches iOS which casts to an empty string. Although it's not very useful without the link.

PR #414, opened 2026-06-09
feat: add logs API (captureLog + logger facade)

turnipdabeets:

Brings PostHog's structured logging to Flutter, matching the iOS, Android, web, and React Native SDKs. App developers can now ship logs from their Flutter app and see them next to their events and session replays.

It's a thin wrapper: Dart `Posthog().captureLog(...)` / `Posthog().logger.{trace…fatal}` forward over the method channel to the native SDKs (iOS `PostHogSDK.captureLog`, Android `PostHog.captureLog`, posthog-js on web), which handle batching, persistence, rate-capping, OTLP transport, and auto-context. Identity/tuning config + a Dart-side `beforeSend` live on `config.logsConfig`. W3C trace fields are optional and pass-through.

Raises the `posthog-android` floor to `3.48.0` (the release with the public `captureLog` trace API).

Closes PostHog/posthog#40528.

- **Unit tests (161):** capture-time gating, `beforeSend` (incl. throwing-hook drops the log), severity→wire mapping per level, trace fields incl. the explicit `traceFlags: 0` case, config serialization, and the web handler (run in CI via `flutter test --platform chrome`).
- **On-device, end-to-end, all three platforms** — captured `captureLog` + `logger.error`, confirmed the OTLP `POST /i/v1/logs → 200` and the records are queryable in PostHog with correct severities, attributes, and native auto-context:
  - **Android** — physical Pixel 4 (Android 13).
  - **iOS** — iPhone 17 simulator (scope `posthog-ios`).
  - **Web** — Chrome via a production `flutter build web` + browser load (`url.full` auto-context; posthog-js path). Note: `flutter run -d chrome` hits an unrelated flutter-tooling DWDS crash in my env, so web was verified via the built app.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/logs-api.md
+++ b/.changeset/logs-api.md
@@ -12,6 +12,8 @@ Posthog().logger.error('payment failed', {'error_code': 'E001'});
 await Posthog().captureLog(body: 'request finished', level: PostHogLogSeverity.warn);
 ```
 
-Levels: `trace`, `debug`, `info`, `warn`, `error`, `fatal`. Configure service identity, redaction (`beforeSend`), and batching/rate-cap tuning on `config.logs` — all optional, with sensible native defaults. Works on iOS, Android, and web.
+Levels: `trace`, `debug`, `info`, `warn`, `error`, `fatal`. Configure service identity, redaction (`beforeSend`), and batching/rate-cap tuning on `config.logsConfig` — all optional, with sensible native defaults. Works on iOS, Android, and web.
+
+Requires `posthog-android` `3.48.0` or newer.
 
 See https://posthog.com/docs/logs for details.
--- a/.changeset/logs-api.md
+++ b/.changeset/logs-api.md
@@ -2,47 +2,16 @@
 "posthog_flutter": minor
 ---
 
-Add structured logging. You can now send logs to PostHog from your Flutter app and see them next to your events and session replays — great for debugging what actually happened in production.
-
-Use `logger` for everyday logging:
+Add structured logging. Send logs to PostHog from your Flutter app and see them next to your events and session replays.
 
 ```dart
 Posthog().logger.info('checkout completed', {'order_id': 'ord_789'});
-Posthog().logger.warn('retrying payment', {'attempt': 2});
 Posthog().logger.error('payment failed', {'error_code': 'E001'});
+
+// Or pick the level at runtime:
+await Posthog().captureLog(body: 'request finished', level: PostHogLogSeverity.warn);
 ```
 
-Levels are `trace`, `debug`, `info`, `warn`, `error`, and `fatal`, and you can attach any JSON-friendly attributes. When you need to pick the level at runtime, use `captureLog` instead:
+Levels: `trace`, `debug`, `info`, `warn`, `error`, `fatal`. Configure service identity, redaction (`beforeSend`), and batching/rate-cap tuning on `config.logs` — all optional, with sensible native defaults. Works on iOS, Android, and web.
 
-```dart
-await Posthog().captureLog(
-  body: 'request finished',
-  level: ok ? PostHogLogSeverity.info : PostHogLogSeverity.error,
-  attributes: {'duration_ms': 142},
-);
-```
-
-Every log is automatically tagged with the current distinct ID, session, screen, app state, and active feature flags, so you get that context for free.
-
-Configure it (everything optional — anything you skip keeps a sensible default) on `config.logs`:
-
-```dart
-config.logs
-  ..serviceName = 'checkout-app'
-  ..environment = 'production'
-  // Redact or drop logs before they leave the device:
-  ..beforeSend = [
-    (record) {
-      record.attributes?.remove('password');
-      return record;
-    },
-  ];
-```
-
-You can also tune batching and rate limiting (`flushInterval`, `flushAt`, `maxBatchSize`, `maxBufferSize`, `rateCapMaxLogs`, `rateCapWindow`) if the defaults don't suit you. Works on iOS, Android, and web.
-
-**Good to know:**
-
-- Distributed tracing: `captureLog` also accepts optional `traceId`, `spanId`, and `traceFlags` to correlate a log with a trace.
-- On Flutter web, set your log options in your own `posthog.init({...})` call (Flutter attaches to your existing posthog-js instance) — your Dart `beforeSend` still runs.
-- Android: needs `posthog-android` `3.48.0` or newer, which this package now requires for you.
+See https://posthog.com/docs/logs for details.
--- a/.changeset/logs-api.md
+++ b/.changeset/logs-api.md
@@ -1,0 +1,48 @@
+---
+"posthog_flutter": minor
+---
+
+Add structured logging. You can now send logs to PostHog from your Flutter app and see them next to your events and session replays — great for debugging what actually happened in production.
+
+Use `logger` for everyday logging:
+
+```dart
+Posthog().logger.info('checkout completed', {'order_id': 'ord_789'});
+Posthog().logger.warn('retrying payment', {'attempt': 2});
+Posthog().logger.error('payment failed', {'error_code': 'E001'});
+```
+
+Levels are `trace`, `debug`, `info`, `warn`, `error`, and `fatal`, and you can attach any JSON-friendly attributes. When you need to pick the level at runtime, use `captureLog` instead:
+
+```dart
+await Posthog().captureLog(
+  body: 'request finished',
+  level: ok ? PostHogLogSeverity.info : PostHogLogSeverity.error,
+  attributes: {'duration_ms': 142},
+);
+```
+
+Every log is automatically tagged with the current distinct ID, session, screen, app state, and active feature flags, so you get that context for free.
+
+Configure it (everything optional — anything you skip keeps a sensible default) on `config.logs`:
+
+```dart
+config.logs
+  ..serviceName = 'checkout-app'
+  ..environment = 'production'
+  // Redact or drop logs before they leave the device:
+  ..beforeSend = [
+    (record) {
+      record.attributes?.remove('password');
+      return record;
+    },
+  ];
+```
+
+You can also tune batching and rate limiting (`flushInterval`, `flushAt`, `maxBatchSize`, `maxBufferSize`, `rateCapMaxLogs`, `rateCapWindow`) if the defaults don't suit you. Works on iOS, Android, and web.
+
+**Good to know:**
+
+- Distributed tracing: `captureLog` also accepts optional `traceId`, `spanId`, and `traceFlags` to correlate a log with a trace.
+- On Flutter web, set your log options in your own `posthog.init({...})` call (Flutter attaches to your existing posthog-js instance) — your Dart `beforeSend` still runs.
+- Android: needs `posthog-android` `3.48.0` or newer, which this package now requires for you.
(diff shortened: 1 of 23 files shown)


posthog_flutter/lib/src/posthog_config.dart
```diff
@@ -284,6 +302,131 @@ class PostHogConfig {
...
+      if (resourceAttributes.isNotEmpty)
+        'resourceAttributes': resourceAttributes,
+      if (flushInterval != null)
+        'flushIntervalSeconds': flushInterval!.inSeconds,
+      if (flushAt != null) 'flushAt': flushAt,
+      if (maxBatchSize != null) 'maxBatchSize': maxBatchSize,
+      if (maxBufferSize != null) 'maxBufferSize': maxBufferSize,
+      if (rateCapMaxLogs != null) 'rateCapMaxLogs': rateCapMaxLogs,
+      if (rateCapWindow != null)
+        'rateCapWindowSeconds': rateCapWindow!.inSeconds,
```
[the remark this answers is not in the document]
turnipdabeets: Addressed Greptile's P2 (sub-second Duration truncating to 0) in https://github.com/PostHog/posthog-flutter/commit/4b5dd66e45f47364a73a1d80b3e0262180ee0065: flushInterval/rateCapWindow now floor at 1s — the smallest the seconds-granular native API can honor — so a tiny window can't silently disable the rate cap. Added a regression test.

.changeset/logs-api.md
```diff
@@ -0,0 +1,48 @@
...
+    },
+  ];
+```
+
+You can also tune batching and rate limiting (`flushInterval`, `flushAt`, `maxBatchSize`, `maxBufferSize`, `rateCapMaxLogs`, `rateCapWindow`) if the defaults don't suit you. Works on iOS, Android, and web.
+
+**Good to know:**
+
+- Distributed tracing: `captureLog` also accepts optional `traceId`, `spanId`, and `traceFlags` to correlate a log with a trace.
+- On Flutter web, set your log options in your own `posthog.init({...})` call (Flutter attaches to your existing posthog-js instance) — your Dart `beforeSend` still runs.
```
marandaneto: not against this, but we could simply link to our docs instead

.changeset/logs-api.md
```diff
@@ -0,0 +1,48 @@
...
+- Android: needs `posthog-android` `3.48.0` or newer, which this package now requires for you.
```
marandaneto: just noise since dependency resolution will force that, its not something the user can do anything about

posthog_flutter/lib/src/utils/before_send.dart
```diff
@@ -0,0 +1,29 @@
...
+) async {
+  try {
+    final result = callback(record);
+    if (result is Future<T?>) {
+      return await result;
+    }
+    return result;
+  } catch (e) {
+    printIfDebug('[PostHog] beforeSend callback threw exception: $e');
+    return record;
```
marandaneto: should we drop or return as is? what do we do in the other SDKs?

posthog_flutter/lib/src/posthog.dart
```diff
@@ -227,6 +231,100 @@ class Posthog {
...
+  /// Per-level logger facade for capturing structured logs.
+  ///
+  /// Each helper delegates to [captureLog] with the matching severity.
+  ///
+  /// **Example:**
+  /// ```dart
+  /// Posthog().logger.info('user signed in', {'method': 'google'});
+  /// Posthog().logger.error('payment failed', {'error_code': 'E001'});
+  /// ```
+  late final PostHogLogger logger = PostHogLogger(
```
marandaneto: why is it late?

posthog_flutter/lib/src/posthog_flutter_web_handler.dart
```diff
@@ -205,6 +207,36 @@ Future<dynamic> handleWebMethodCall(MethodCall call) async {
...
+    case 'captureLog':
+      final body = args['body'] as String;
+      final level = args['level'] as String? ?? 'info';
+      final attributes = safeMapConversion(args['attributes']);
+      final traceId = args['traceId'] as String?;
+      final spanId = args['spanId'] as String?;
+      final traceFlags = args['traceFlags'] as int?;
+
+      // posthog-js captureLog options. See:
+      // https://github.com/PostHog/posthog-js/blob/main/packages/types/src/capture-log.ts
```
marandaneto: set it as a permalink or just to the docs
this link will be outdated at some point

posthog_flutter/lib/src/posthog_config.dart
```diff
@@ -284,10 +302,142 @@ class PostHogConfig {
       'dataMode': dataMode.name,
       'sessionReplayConfig': sessionReplayConfig.toMap(),
       'errorTrackingConfig': errorTrackingConfig.toMap(),
+      'logs': logs.toMap(),
     };
   }
 }
 
+/// Configuration for the logs subsystem.
```
marandaneto, line 310: this requires manual init to work right, similar to session replay and error tracking?

marandaneto, line 310: once you call `captureLog` all logs are captured, its not respecting the remote config right (if disabled), or is it on the mobile sdk side? we have to be clearer about that

marandaneto:
> Manually on a physical Android device (Pixel 4)

also make sure to test on an iOS emulator and on the browser, its unclear if you did

marandaneto:
looks good, left a few comments but its nearly there

marandaneto:
@turnipdabeets you can update your skill to remove the `release` label, its not used anymore

ioannisj:
@turnipdabeets will have a look later today at this

posthog_flutter/lib/src/posthog_config.dart
turnipdabeets: no manual init, this is manual capture so the develop manually sends these. On auto capture we'll have to add an init and respect remote config.

turnipdabeets: correct, I'll add to documentation to be clear then?

posthog_flutter/lib/src/utils/before_send.dart
turnipdabeets: events continue, dont drop. I think Android is the only one that drops logs. Maybe we should make these consistent. Dropping prevents leaking anything, but continue might be the expectation. Do you have an opinion on this @marandaneto ?

posthog_flutter/lib/src/posthog.dart
turnipdabeets: It was late because the field initializer closed over this.captureLog (an instance method), and Dart doesn't allow a non-late instance field initializer to reference this, but I'm switching to a lazy init since that might be more clear.

posthog_flutter/lib/src/utils/before_send.dart
marandaneto: mm, i imagine that if you are setting up a callback maybe you are doing PII
if you had an error, its better to drop than returning the event, you might be leaking PII
ofc someone could just be enriching the data, so better safe than sorry imo

marandaneto: log out and drop would be my call, so they learn with the diagnostic logging at least

posthog_flutter/lib/src/posthog_config.dart
marandaneto: also worth noting in dartdoc

posthog_flutter/lib/src/utils/before_send.dart
turnipdabeets: ok went with that, addressed JS to match too https://github.com/PostHog/posthog-js/pull/3799

posthog_flutter/lib/src/posthog_config.dart
```diff
@@ -284,10 +302,142 @@ class PostHogConfig {
...
+  /// ```
+  ///
+  /// **Note:**
+  /// - Runs in Dart on all platforms — it is intentionally not forwarded to the
+  ///   native SDKs' own `beforeSend`. Dart callbacks cannot cross the platform
+  ///   channel, and running it here gives identical behavior everywhere
+  ///   (including web). This mirrors the event [PostHogConfig.beforeSend].
+  /// - Callbacks can be synchronous or asynchronous (via
+  ///   `FutureOr<PostHogLogRecord?>`).
+  /// - Exceptions in a callback skip that callback and continue with the next.
```
ioannisj: Needs to be update to match what was done here https://github.com/PostHog/posthog-flutter/pull/414#discussion_r3386864359

posthog_flutter/lib/src/posthog.dart
```diff
@@ -227,6 +231,110 @@ class Posthog {
...
+    int? traceFlags,
+  }) async {
+    if (body.trim().isEmpty) {
+      return;
+    }
+
+    var record = PostHogLogRecord(
+      body: body,
+      level: level,
+      attributes: attributes,
```
ioannisj: We probably need a copy of this since we pass it back into beforeSend hook and can be mutated. I think this means that currently we are mutating the original map?

posthog_flutter/lib/src/posthog_config.dart
```diff
@@ -157,6 +171,10 @@ class PostHogConfig {
   /// Configuration for error tracking and exception capture.
   final errorTrackingConfig = PostHogErrorTrackingConfig();
 
+  /// Configuration for the logs subsystem (`Posthog().captureLog()` and the
+  /// `Posthog().logger` facade).
+  final logs = PostHogLogsConfig();
```
ioannisj: nit: other configs have `_Config` prefix, so should this follow? e.g logsConfig. Raising now since it can't change after we ship

posthog_flutter/test/posthog_flutter_web_handler_test.dart
```diff
@@ -0,0 +1,80 @@
+@TestOn('browser')
```
ioannisj, line 1: Do we run this in the CI?

posthog_flutter/darwin/posthog_flutter/Sources/posthog_flutter/PosthogFlutterPlugin.swift
```diff
@@ -719,6 +756,50 @@ extension PosthogFlutterPlugin {
...
+            result(nil)
+        } else {
+            _badArgumentError(result)
+        }
+    }
+
+    // Maps the lowercase wire level to PostHogLogSeverity using only public enum
+    // cases (the SDK's `from(name:)` is internal). Unknown levels fall back to .info.
+    private func severityFromString(_ level: String) -> PostHogLogSeverity {
+        switch level {
```
ioannisj: I think this reaches lowercased from flutter side, but should we lowercase anyway here to match android plugin is doing?

.changeset/logs-api.md
```diff
@@ -0,0 +1,17 @@
+---
+"posthog_flutter": minor
+---
+
+Add structured logging. Send logs to PostHog from your Flutter app and see them next to your events and session replays.
```
ioannisj, line 5: Let's maybe also mention the posthog-android bump to 3.48.0?

ioannisj (commented):
Looks good, left some comments as well, minor for the most part. I'll leave the approval on Manoel since there are some of his threads still open

ioannisj:
Let's also close https://github.com/PostHog/posthog/issues/40528 once this is merged

.changeset/logs-api.md
turnipdabeets: added

posthog_flutter/lib/src/posthog_config.dart
turnipdabeets: good point, renamed

posthog_flutter/lib/src/posthog.dart
turnipdabeets: copying now

posthog_flutter/test/posthog_flutter_web_handler_test.dart
turnipdabeets: it didn't but it should now

posthog_flutter/lib/src/posthog_config.dart
turnipdabeets: updated

ioannisj (approved):
LG for me

PR #442, opened 2026-06-16
feat: add addExceptionStep API for error tracking

ioannisj:

Adds **exception steps** to error tracking for the Flutter SDK.

Exception steps are breadcrumb-style context records that accumulate over time and attach to every captured `$exception` as `$exception_steps`, giving the error-tracking UI a timeline of what happened right before the exception.

```dart
Posthog().addExceptionStep('User tapped Checkout', properties: {'screen': 'cart'});
```

> [!NOTE]
> Requires the released native SDKs that carry the feature: **posthog-android ≥ 3.51.0** and **posthog-ios ≥ 3.61.0**. The dependency pins are bumped accordingly in this PR.

### What's included

- **`addExceptionStep(message, {properties})`** public API on `Posthog`, forwarding to the native layer (where the buffer, byte-budget eviction, serialization, and crash-durable persistence live).
- **`PostHogExceptionStepsConfig`** on `errorTrackingConfig.exceptionSteps` — `enabled` (default `true`) and `maxBytes` (default `32768`) — forwarded to the native config at `setup`.
- **Native wiring** in the Android (Kotlin) and iOS (Swift) plugins: method-channel handlers + config plumbing. Steps attach to Flutter-captured exceptions because they flow through the native `$exception` capture path.
- **Dependency bumps**: `posthog-android` → `[3.51.0,4.0.0)`; `posthog-ios` → `3.61.0..<4.0.0` (SPM + podspec).
- **Web**: forwarded to posthog-js's `addExceptionStep` (guarded for older builds). Note: on web, steps attach to exceptions captured by posthog-js, not to those captured via `Posthog().captureException()` — documented on the API.
- **Example app**: a dedicated "Exception Steps" screen for manual testing (add single/batch steps with mixed property types, high-frequency flood, and trigger handled / unhandled / native-crash exceptions).

- Unit tests for the public API (forwarding + empty-message no-op), the method-channel IO layer (message + normalized properties), and the web handler. `flutter analyze` clean; all Dart + web tests pass.
- Verified end-to-end on the iOS Simulator and Android emulator via the example screen, building against the released native SDKs (iOS 3.61.0 / Android 3.51.0): confirmed the captured `$exception` carries `$exception_steps` with correctly normalized properties (nested maps, lists, numbers, bools, `DateTime`), across the handled, unhandled, and native-crash paths.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/posthog_flutter/lib/src/posthog.dart
+++ b/posthog_flutter/lib/src/posthog.dart
@@ -675,9 +675,10 @@ class Posthog {
   /// next launch.
   ///
   /// The [message] is a short, non-empty description of what happened; an empty
-  /// message is ignored. The optional [properties] are additional context. The
-  /// reserved keys `$message` and `$timestamp` are stripped — the SDK sets the
-  /// canonical values, including a timestamp captured when the step is recorded.
+  /// or whitespace-only message is ignored. The optional [properties] are
+  /// additional context. The reserved keys `$message` and `$timestamp` are
+  /// stripped — the SDK sets the canonical values, including a timestamp
+  /// captured when the step is recorded.
   ///
   /// Recording never throws into your app and does not block the caller.
   ///
@@ -697,7 +698,7 @@ class Posthog {
     String message, {
     Map<String, Object>? properties,
   }) {
-    if (message.isEmpty) {
+    if (message.trim().isEmpty) {
       debugPrint('[PostHog] addExceptionStep called with an empty message.');
       return Future<void>.value();
     }
--- a/posthog_flutter/lib/src/posthog.dart
+++ b/posthog_flutter/lib/src/posthog.dart
@@ -701,6 +701,12 @@ class Posthog {
       debugPrint('[PostHog] addExceptionStep called with an empty message.');
       return Future<void>.value();
     }
+    // Honor the documented no-op contract on every platform: native enforces
+    // `enabled` via the config forwarded at setup, but on web `setup` doesn't
+    // push it to posthog-js, so guard here too.
+    if (_config?.errorTrackingConfig.exceptionSteps.enabled == false) {
+      return Future<void>.value();
+    }
     return _posthog.addExceptionStep(message, properties: properties);
   }
 
--- a/posthog_flutter/lib/src/posthog.dart
+++ b/posthog_flutter/lib/src/posthog.dart
@@ -661,6 +661,49 @@ class Posthog {
     );
   }
 
+  /// Records an exception step (breadcrumb-style context record).
+  ///
+  /// Steps accumulate in a rolling, byte-bounded buffer and are attached to
+  /// every captured `$exception` event as `$exception_steps`, giving the
+  /// PostHog error-tracking UI a timeline of recent activity leading up to each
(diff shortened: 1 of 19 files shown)


posthog_flutter/lib/src/posthog.dart
```diff
@@ -661,6 +661,49 @@ class Posthog {
...
+  /// ```
+  Future<void> addExceptionStep(
+    String message, {
+    Map<String, Object>? properties,
+  }) {
+    if (message.isEmpty) {
+      debugPrint('[PostHog] addExceptionStep called with an empty message.');
+      return Future<void>.value();
+    }
+    return _posthog.addExceptionStep(message, properties: properties);
```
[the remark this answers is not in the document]
ioannisj: Fixed

dustinbyrne (commented):
change looks good overall - just one issue we should fix

posthog_flutter/lib/src/posthog_flutter_web_handler.dart
```diff
@@ -348,6 +350,22 @@ Future<dynamic> handleWebMethodCall(MethodCall call) async {
...
       break;
+    case 'addExceptionStep':
+      final message = args['message'] as String;
+      final properties = safeMapConversion(args['properties']);
+
+      try {
+        posthog?.addExceptionStep(
+          stringToJSAny(message),
+          properties.isNotEmpty ? mapToJSAny(properties) : null,
+        );
```
dustinbyrne, line 381: there's no override in `PosthogFlutterWeb`, so this won't be reached and calling `addExceptionStep` on web will raise an exception

posthog_flutter/lib/src/posthog_config.dart
```diff
@@ -644,6 +650,48 @@ class PostHogErrorTrackingConfig {
...
+
+  /// Total UTF-8 byte budget for the rolling step buffer.
+  ///
+  /// When adding a step would exceed the budget, the oldest steps are evicted
+  /// until the total fits. A single step larger than the budget is rejected
+  /// outright. Defaults to `32768` (32 KiB).
+  var maxBytes = 32768;
+
+  /// Converts this configuration to a platform-channel map.
+  Map<String, dynamic> toMap() {
```
marandaneto: always use object isntead of dynamic
i think there was a linter enabled for that, pls check other places

posthog_flutter/lib/src/posthog_flutter_web_handler.dart
marandaneto: yep, and pls test it

marandaneto (approved):
approving to unblock but 2 legit comments

posthog_flutter/lib/src/posthog_flutter_web_handler.dart
ioannisj: Right, fixed. I also need to write the exception capture through `captureException()` instead of `capture()` because js lib only attaches exception steps through that method. Testing now, this was an omission from my part, thanx!

ioannisj: Addressed and tested, web captures exception steps now

PR #446, opened 2026-06-23
chore: update Flutter PopScope callback

marandaneto:
## :bulb: Motivation and Context

Flutter 3.27 is now the minimum supported version, so survey UI code can use the newer PopScope callback instead of the deprecated API. This also removes a couple of stale TODO comments.

## :green_heart: How did you test it?

- `make checkApiDart`
- `cd posthog_flutter && dart analyze lib/posthog_flutter_web.dart lib/src/posthog_config.dart lib/src/surveys/widgets/survey_bottom_sheet.dart`

## :pencil: Checklist

- [x] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [ ] I updated the docs if needed.
- [x] No breaking change or entry added to the changelog.

### If releasing new changes

- [ ] Ran `pnpm changeset` to generate a changeset file

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

Updated the survey bottom sheet to use `onPopInvokedWithResult`, which is available with the current Flutter minimum, and removed stale TODO notes in the changed files. The branch was merged with the latest `main`; the public API snapshot conflict was resolved by keeping `main`'s version-independent snapshot format.

PR #447, opened 2026-06-23
chore: template public API package version

marandaneto:

The Dart public API snapshot currently includes `packageApi.packageVersion`, so version-only release bumps can fail unrelated PR checks when the checked-in snapshot contains an older package version.

This normalizes the package version in the generated public API snapshot to the stable `<version>` template, matching the intent that public API checks only fail on API shape changes.

```bash
scripts/check-api-dart.sh
```

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with Pi coding agent. The public API check script now replaces the generated `packageApi.packageVersion` with `<version>` before comparing or updating the snapshot, and the checked-in snapshot was updated to use that template. No changeset was added because this is repository tooling/test infrastructure only and does not change the published package behavior.

### DIFF
--- a/api/posthog_flutter.api.json
+++ b/api/posthog_flutter.api.json
@@ -4619,7 +4619,7 @@
             }
         ],
         "packageName": "posthog_flutter",
-        "packageVersion": "5.27.0",
+        "packageVersion": "<version>",
         "sdkType": "flutter",
         "semantics": [
             "containsPlatformConstraints"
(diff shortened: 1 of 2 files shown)


ioannisj (approved):
LG, will probably have a merge conflict with https://github.com/PostHog/posthog-flutter/pull/446

PR #453, opened 2026-06-25
feat: capture platform views in session replay screenshots

turnipdabeets:

Platform views (WebView, Google Maps, etc.) rendered by Flutter are composited outside the Flutter layer tree. Without explicit handling, session replay either silently skips them (leaving blank gaps) or could inadvertently capture sensitive content. This PR adds first-class support: views are masked by default (black box), with opt-in capture on a per-view or global basis.

Closes #393.
Addresses only part 1 of #151

Note I'll address fully native views in a separate PR, that will require iOS and android SDK work as well.

- iOS simulator (iPhone 17, iOS 26.4): ran all 8 example app test flows — masked views appear as black boxes in PostHog replay, captured WebViews show actual content
- Android device (Pixel 4, Android 13): same flows verified
- flutter test in posthog_flutter/ — all tests pass
- Confirmed 5-second timeout fires correctly when native channel does not respond

See videos:

Android: https://us.posthog.com/shared/AYnlFG5M-DGczYUYyn27g1EC55HvDg?t=1
https://us.posthog.com/shared/kWlqsDn5LUynHgynC0Mp1t_wnAfK5w?t=2

iOS: https://us.posthog.com/shared/2G_GNlw70uTOwJQOdVcZFY_uiDasTA?t=5

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Authored with Claude Code (claude-sonnet-4-6). Anna Garcia directed the work throughout.

Key decisions made during the session:

- **iOS: no drawHierarchy fallback** — when no WKWebView is found for the captured rect, we return nil rather than falling back to drawHierarchy over the full window. The fallback would composite any masked CALayer-backed platform view in the crop region and leak it into replay, defeating the privacy guarantee.
- **Android: async bitmap compression** — PixelCopy runs on the main thread; bitmap compression was moved to a single-thread executor to avoid blocking the UI thread during screenshot capture.
- **5-second timeout on captureNativeScreenshot** — if the native side never calls result(), the Dart _isCapturing flag would stay true permanently and silently kill replay. The timeout degrades gracefully to null (frame skipped) rather than freezing.
- **hasCapturedPlatformViews gate** — scheduleFrame() is only forced when at least one platform view was actually captured, avoiding unnecessary redraws on screens with only masked views.

### DIFF
--- a/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
+++ b/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
@@ -5,20 +5,20 @@ import android.content.ActivityNotFoundException
 import android.content.Context
 import android.content.Intent
 import android.graphics.Bitmap
-import android.graphics.PorterDuff
-import android.graphics.PorterDuffXfermode
 import android.graphics.Rect
 import android.net.Uri
-import android.os.Build
 import android.os.Bundle
+import android.os.Build
 import android.os.Handler
+import androidx.annotation.RequiresApi
 import android.os.Looper
-import android.util.Log
+import android.graphics.PorterDuff
+import android.graphics.PorterDuffXfermode
 import android.view.PixelCopy
 import android.view.SurfaceView
 import android.view.View
 import android.view.ViewGroup
-import androidx.annotation.RequiresApi
+import android.util.Log
 import com.posthog.PersonProfiles
 import com.posthog.PostHog
 import com.posthog.PostHogConfig
@@ -27,9 +27,9 @@ import com.posthog.android.PostHogAndroid
 import com.posthog.android.PostHogAndroidConfig
 import com.posthog.android.internal.getApplicationInfo
 import com.posthog.logs.PostHogLogSeverity
-import io.flutter.embedding.engine.plugins.FlutterPlugin
 import io.flutter.embedding.engine.plugins.activity.ActivityAware
 import io.flutter.embedding.engine.plugins.activity.ActivityPluginBinding
+import io.flutter.embedding.engine.plugins.FlutterPlugin
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
 import io.flutter.plugin.common.MethodChannel.MethodCallHandler
@@ -258,6 +258,10 @@ class PosthogFlutterPlugin :
 
             "captureNativeScreenshot" -> {
                 handleCaptureNativeScreenshot(call, result)
+            }
+
+            "captureNativeScreenshots" -> {
+                handleCaptureNativeScreenshots(call, result)
             }
 
             "isSessionReplayActive" -> {
@@ -542,53 +546,74 @@ class PosthogFlutterPlugin :
         result: Result,
     ) {
         try {
-            val currentActivity = activity
-            if (currentActivity == null) {
-                result.success(null)
-                return
-            }
-
+            val currentActivity = activity ?: run { result.success(null); return }
             val x = call.argument<Int>("x") ?: 0
             val y = call.argument<Int>("y") ?: 0
             val width = call.argument<Int>("width") ?: 0
             val height = call.argument<Int>("height") ?: 0
-
             if (width <= 0 || height <= 0) {
                 result.error("INVALID_ARGUMENT", "Width or height is 0", null)
                 return
             }
-
-            captureNativeScreenshot(
-                activity = currentActivity,
-                x = x,
-                y = y,
-                width = width,
-                height = height,
-                result = result,
-            )
+            captureOneNative(currentActivity, x, y, width, height) { bytes ->
+                result.success(bytes)
+            }
         } catch (e: Throwable) {
             result.error("PosthogFlutterException", e.localizedMessage, null)
         }
     }
 
-    private fun captureNativeScreenshot(
+    private fun handleCaptureNativeScreenshots(
+        call: MethodCall,
+        result: Result,
+    ) {
+        try {
+            val currentActivity = activity ?: run { result.success(emptyList<ByteArray?>()); return }
+            @Suppress("UNCHECKED_CAST")
+            val views = call.argument<List<Map<String, Int>>>("views") ?: emptyList()
+            if (views.isEmpty()) {
+                result.success(emptyList<ByteArray?>())
+                return
+            }
+            val results = ArrayList<ByteArray?>(views.size)
+            fun captureNext(index: Int) {
+                if (index >= views.size) {
+                    result.success(results)
+                    return
+                }
+                val v = views[index]
+                val x = v["x"] ?: 0
+                val y = v["y"] ?: 0
+                val w = v["width"] ?: 0
+                val h = v["height"] ?: 0
+                captureOneNative(currentActivity, x, y, w, h) { bytes ->
+                    results.add(bytes)
+                    captureNext(index + 1)
+                }
+            }
+            captureNext(0)
+        } catch (e: Throwable) {
+            result.error("PosthogFlutterException", e.localizedMessage, null)
+        }
+    }
+
+    private fun captureOneNative(
         activity: Activity,
         x: Int,
         y: Int,
         width: Int,
         height: Int,
-        result: Result,
+        onResult: (ByteArray?) -> Unit,
     ) {
-        val contentView =
-            activity.findViewById<View>(android.R.id.content) ?: run {
-                result.success(null)
-                return
-            }
+        val contentView = activity.findViewById<View>(android.R.id.content) ?: run {
+            onResult(null)
+            return
+        }
 
         val contentWidthPx = contentView.width
         val contentHeightPx = contentView.height
         if (contentWidthPx <= 0 || contentHeightPx <= 0) {
-            result.success(null)
+            onResult(null)
             return
         }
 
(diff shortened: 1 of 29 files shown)


posthog_flutter/lib/src/replay/screenshot/screenshot_capturer.dart
```diff
@@ -88,6 +97,139 @@ class ScreenshotCapturer {
...
+    try {
+      final codec = await ui.instantiateImageCodec(bytes);
+      final frame = await codec.getNextFrame();
+      codec.dispose();
+      return frame.image;
+    } catch (e) {
+      printIfDebug('Error decoding image bytes: $e');
+      return null;
+    }
+  }
```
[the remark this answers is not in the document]
turnipdabeets: Fixed in e491499 — moved `codec.dispose()` into a `finally` block so it runs on both the happy path and when `getNextFrame()` throws.

posthog_flutter/lib/src/replay/mask/posthog_platform_view.dart
```diff
@@ -0,0 +1,48 @@
...
+/// Controls how a platform view is handled in session replay.
+enum PostHogPlatformViewPrivacy {
+  /// Mask the view with a black rectangle (default).
+  mask,
+
+  /// Reveal the view content in session replay.
+  ///
+  /// On Android texture mode the content is already in the Flutter render base,
+  /// so no extra capture work is done. On iOS and hybrid composition the
+  /// compositor fills the transparent hole via a native capture + dstOver.
```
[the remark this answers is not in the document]
turnipdabeets: Fixed in e491499 — doc comment updated to `srcOver` to match the actual blend mode used in `_compositeRevealedView`.

posthog_flutter/lib/src/replay/screenshot/screenshot_capturer.dart
```diff
@@ -227,12 +384,18 @@ class ScreenshotCapturer {
...
         }
 
-        final currentHash = _computeImageHash(imageBytes);
+        final preMaskHash = _computeImageHash(imageBytes);
         imageBytes = null;
 
-        if (currentHash == statusView.imageBytesHash) {
+        final pvRects = replayConfig.maskAllPlatformViews
+            ? _collectPlatformViewRects()
+            : null;
```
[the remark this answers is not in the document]
turnipdabeets: Fixed in 8cb2b93.

`_collectPlatformViewRects` now always walks the element tree regardless of `maskAllPlatformViews`. The global setting just determines the *default* policy passed into the walk:

- `maskAllPlatformViews = true` → default is `.mask` (per-view `.capture` overrides reveal)
- `maskAllPlatformViews = false` → default is `.capture` (per-view `.mask` overrides now correctly mask)

A `PostHogPlatformView(privacy: .mask)` wrapper around a sensitive view will be respected in both modes.

posthog_flutter/lib/src/replay/native_communicator.dart
```diff
@@ -51,4 +51,31 @@ class NativeCommunicator {
       return false;
     }
   }
+
+  Future<Uint8List?> captureNativeScreenshot({
```
ioannisj: I think this is worth discussing first before patching around it. 

Today this is one call per captured view, per frame, each carrying a png data (could be low/high MB range?) Can we maybe design this to be 1 call per snapshot at least?

posthog_flutter/lib/src/replay/native_communicator.dart
```diff
@@ -51,4 +51,31 @@ class NativeCommunicator {
...
+    required int width,
+    required int height,
+  }) async {
+    if (kIsWeb) {
+      return null;
+    }
+    try {
+      final bytes = await _channel.invokeMethod<Uint8List>(
+        'captureNativeScreenshot',
+        {'x': x, 'y': y, 'width': width, 'height': height},
```
ioannisj: Instead of geometry, can we be passing and threading flutter view id here? Then native side can just lookup per id?

posthog_flutter/lib/src/replay/screenshot/screenshot_capturer.dart
```diff
@@ -277,6 +440,18 @@ class ScreenshotCapturer {
...
         }
 
+        if (pvRects.masked.isNotEmpty) {
+          _imageMaskPainter.drawMaskedImage(
+            canvas,
+            pvRects.masked,
+            pixelRatio,
+          );
+        }
+        for (final capturedRect in pvRects.captured) {
```
ioannisj: see https://github.com/PostHog/posthog-flutter/pull/453/changes#r3497391223

this essentially becomes `final results = await _nativeCommunicator.captureNativeScreenshots()` with one await, then we loop through the return list and composite

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -495,6 +537,347 @@ class PosthogFlutterPlugin :
...
+                    croppedBitmap.recycle()
+                }
+            }
+
+        compressBitmapToPngAsync(outputBitmap, result)
+    }
+
+    private fun bitmapToPng(bitmap: Bitmap): ByteArray {
+        val outputStream = ByteArrayOutputStream()
+        bitmap.compress(Bitmap.CompressFormat.PNG, 100, outputStream)
```
ioannisj: Wonder if there is a way to avoid encoding/decoding PNG-100 during the native hop from dart? Batching will also help. 

Maybe for this path we encode into something cheaper since it's just capturing the platform view snapshot?

posthog_flutter/lib/src/replay/native_communicator.dart
turnipdabeets: The challenge is on the native side: we'd need platform-specific lookup logic in each plugin, and the API for that (FlutterView.getPlatformViewsController() on Android) is internal. Geometry lets us use PixelCopy / WKWebView.takeSnapshot uniformly across all composition modes without touching Flutter internals.

Open to revisiting if you have a native lookup path in mind — what API on the Android or iOS side were you thinking of?

posthog_flutter/lib/src/replay/native_communicator.dart
turnipdabeets: batched in  ae7a7e3

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
turnipdabeets: Updated — raw RGBA (skipping PNG entirely). Native returns raw pixel bytes, Dart decodes with decodeImageFromPixels. No codec round-trip.

posthog_flutter/lib/src/replay/native_communicator.dart
ioannisj: Was thinking about `getFlutterEngine().getPlatformViewsControllerDelegator().getPlatformViewById()` for Android but you are right, this is not symmetric between iOS and Android. So seems possible on Android, but not on iOS. I'd rather have symmetric implementations between the two so we can drop the idea

ioannisj (approved):

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -495,6 +540,376 @@ class PosthogFlutterPlugin :
...
+        if (destX < -tolerance || destY < -tolerance ||
+            destX + svLogW > destBitmap.width + tolerance ||
+            destY + svLogH > destBitmap.height + tolerance
+        ) {
+            advance()
+            return
+        }
+
+        val svBitmap = Bitmap.createBitmap(svLogW, svLogH, Bitmap.Config.ARGB_8888)
+        PixelCopy.request(
```
ioannisj: Could this throw, and should we wrap?

turnipdabeets: Yes —  throws  if the  detaches between the  guard and the call. Since this executes inside a  callback, the outer  in  won't catch it. Fixed in 2804b01 — wrapped in , recycling the bitmap and calling  on failure.

marandaneto:
@ioannisj @turnipdabeets i think lots of code here is duplicated and already exists in the native sdks such as pixel bitmap taking etc
a good follow up is to expose those in the native sdks and reuse here, this helps with reducing duplication/source of bugs and bundle size in general

posthog_flutter/lib/src/replay/mask/posthog_platform_view.dart
```diff
@@ -0,0 +1,57 @@
...
+  /// compositor fills the transparent hole via a native capture + srcOver.
+  ///
+  /// **iOS note:** only [WKWebView]-backed platform views (e.g. `webview_flutter`
+  /// in the default web-view mode) are actually captured on iOS. All other
+  /// platform view types (Google Maps, ARKit, camera previews, etc.) are masked
+  /// regardless of this policy, because the iOS compositor cannot safely
+  /// snapshot arbitrary CALayer-backed views without leaking unmasked content.
+  ///
+  /// **macOS note:** capture is not supported on macOS; all platform views are
+  /// masked regardless of this policy.
```
marandaneto, line 21: we dont support macos, so this is misleading

PR #466, opened 2026-07-02
feat: capture exception cause chains

hpouillot:
`captureException` previously emitted only the outer error in `$exception_list`, losing root-cause context for wrapped async errors and custom exceptions with a `cause` chain.

This adds cause-chain extraction for `AsyncError`, `ParallelWaitError`, and exceptions exposing a `cause` getter. Causes are emitted outermost-first with cycle protection and a maximum chain length.

- `flutter test posthog_flutter/test/dart_exception_processor_test.dart`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi coding agent helped narrow the branch to the cause-chain work only, removing unrelated error-tracking signal changes from the branch diff. The implementation keeps cause stack traces only when available on the cause itself, avoids generating synthetic stacks for causes, and includes tests for async errors, parallel waits, duck-typed causes, cycles, and chain length capping.

### DIFF
--- a/.changeset/error-tracking-signal-improvements.md
+++ b/.changeset/error-tracking-signal-improvements.md
@@ -2,4 +2,4 @@
 "posthog_flutter": minor
 ---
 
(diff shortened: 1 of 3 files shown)


cat-ph (approved):
nice! 🚀

turnipdabeets (approved):

PR #473, opened 2026-07-13
feat(replay): capture native screens that cover Flutter

### LINKED ISSUE #151: Session Recordings: Platform Views (Native Views)

turnipdabeets:

Fixes #151. Session replay froze on the last Flutter frame whenever a fully native screen (a paywall, login, or any view controller/activity presented outside the Flutter engine) covered the app — and touches kept recording onto that stale frame, so replays showed taps landing on the wrong screen.

This adds an out-of-engine occlusion detector and hands the covered stretch to the native SDK's session-replay capture, so the native screen is actually recorded. Opt-in via `PostHogSessionReplayConfig.captureNativeScreens` (default `false` — behavior is unchanged unless enabled).

## :hammer: Changes

- **Occlusion detection**: plugin-side 1 Hz main-thread tickers (independent of Flutter's frame lifecycle, which can pause under a native cover) push episode transitions to Dart.
  - iOS: opaque-coverage walk of the presentation chain + foreign-key-window detection (Superwall-style own-window presentation), scene-scoped, frozen while the app is inactive.
  - Android: host-activity-STOPPED + own-process-activity-RESUMED lifecycle signal.
- **Capture handoff**: while occluded, Flutter capture is suppressed and the native SDK snapshots the covering screen on the detector's cadence (episode-scoped `enableNativeBridge` handshake); falls back to a single black placeholder frame if the native side declines or repeatedly fails before the episode's first frame.
- **Masking parity**: `maskAllTexts`/`maskAllImages` are forwarded to the native SDKs, so captured native frames honor them (with the defaults, all native text and images are masked).
- **What is not captured** (documented on the flag and in the changeset): partial-height sheets (Apple Pay, share sheet), other-process content, Android dialogs/Custom Tabs, and iOS covers without an opaque background — those keep the previous behavior.
- Example app: "Out-of-engine native screens" cards (capture on, capture off, iOS own-window Superwall-style).

- A/B replay verification on iOS (simulator) and Android (Pixel 6 emulator) for capture on/off and masked/revealed configs: native paywall captured with the flag on; pre-feature behavior with it off.
- Dart: episode protocol suite (`native_occlusion_episode_test.dart` — placeholder meta/full pairing, stale-episode drops, episode-end frame re-arm, handshake episode id), `native_communicator_test.dart`, `change_detector_test.dart` (forced-frame suppression), config plumbing in `posthog_test.dart`. 209 tests green, analyze clean.
- Kotlin: `enableNativeBridge` gate tests (declines when not occluded / stale episode; declines never disarm). Fixing the unit-test setup was part of this (`returnDefaultValues`; the suite previously failed at plugin construction).
- Known gap (disclosed): the native detector state machines (iOS 95%-coverage test and strike/demotion, Android teardown races) have no native test harness in this repo — verified by compile + build + review; a Robolectric/instrumentation harness is a follow-up.

https://us.posthog.com/shared/flJEuEYM_9BaQaM-5J0gGJobJm7tJw?t=37

## :warning: Blocked on native releases — draft until then

The capture handoff calls new internal hooks that are not in released SDKs yet:

- posthog-ios: PostHog/posthog-ios#714 (`@_spi(PostHogInternal) captureSessionReplaySnapshot`)
- posthog-android: PostHog/posthog-android#613 (`@PostHogInternalReplayApi captureSessionReplaySnapshot`)

This branch points at the current released dependency ranges, so native CI builds will fail until those ship and the floors here are raised to the releases containing the hook. Do not merge before that.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

- Implemented with Claude Code (Opus 4.8 and Fable 5 sessions) driven by @turnipdabeets, together with the two native SPI hooks it consumes.
- Naming and API shape were deliberated explicitly: `captureNativeScreens` (outcome-named, matching the SDK's `capture*` family) over the implementation-named `bridgeNativeScreens`; kept separate from `maskAllPlatformViews` because the flags govern different sources (embedded views inside the screenshot vs screens invisible to it) and merging them would silently expand capture for existing opt-outs.
- A multi-agent review pass (Dart/iOS/Android/docs reviewers + adversarial verification) preceded this PR; confirmed findings were fixed here (episode-scoped handshake, episode-end frame re-arm, iOS foreign-window opacity/scene checks, inactive-app tick freeze, Flutter-window resolution for platform-view capture, Android detach-safe callbacks) and the docs were rewritten to state the detection envelope honestly.


posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -499,29 +596,306 @@ class PosthogFlutterPlugin :
...
+                }
+            }
+            val scheduled =
+                try {
+                    replayIntegration()?.captureSessionReplaySnapshot(
+                        activity?.window?.peekDecorView(),
+                        isFirst,
+                        { isOccluded && bridgeEnabled && occlusionEpisode == episode },
+                    ) { delivered -> postResult(delivered) } ?: false
+                } catch (e: Throwable) {
```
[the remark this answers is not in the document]
turnipdabeets: The Flutter host's decor view is passed as `excludeView` — the view the SDK must *never* capture — not as the capture target. On the native side, `captureSessionReplaySnapshot` independently picks the topmost `TYPE_BASE_APPLICATION` decor from `Curtains.rootViews` (the resumed covering activity) and refuses to schedule at all if that resolves to the excluded Flutter view. So the covering activity is what gets snapshotted; the stopped host is only the exclusion guard.

posthog_flutter/lib/src/posthog_widget.dart
```diff
@@ -32,6 +34,19 @@ class PostHogWidgetState extends State<PostHogWidget> {
   bool _isCapturing = false;
   bool _disposed = false;
 
+  /// Whether a substitute source (bridge or placeholder) owns the current
+  /// episode. Not the same as "occluded": with the bridge off, occlusion is
+  /// ignored and Flutter capture keeps running.
+  bool _suppressFlutterCapture = false;
+
+  @visibleForTesting
+  bool get debugFlutterCaptureSuppressed => _suppressFlutterCapture;
```
ioannisj, line 43: heads up, the public-api CI job (make checkApiDart) is going to fail on this. api/posthog_flutter.api.json doesn't have PostHogSessionReplayConfig.captureNativeScreens, or the two new public members on the exported PostHogWidgetState (debugFlutterCaptureSuppressed and the static episodeStillCurrent). make updateApiDart sorts the captureNativeScreens diff. the other two read like debug/test hooks though, is there a reason they need to be on the exported class? I'd rather not commit to them as public API, could pull them into a non-exported test-support symbol in lib/src. wdyt?

posthog_flutter/lib/src/posthog_flutter_io.dart
```diff
@@ -81,6 +82,16 @@ class PosthogFlutterIO extends PosthogFlutterPlatformInterface {
       case 'onFeatureFlagsCallback':
         _onFeatureFlagsCallback?.call();
         break;
+      case 'onNativeOcclusionChanged':
```
ioannisj, line 85: is it easy to add a test for onNativeOcclusionChanged here? it's the only thing turning a native occlusion push into the PostHogInternalEvents state the rest of the feature keys off, and the episode suite drives PosthogFlutterPlatformFake so the real PosthogFlutterIO handler never actually runs. not asking for a big harness, just wondering if a quick posthog_flutter_io_test.dart case pushing a real call through TestDefaultBinaryMessengerBinding is cheap to drop in. wdyt?

posthog_flutter/lib/src/posthog_widget.dart
```diff
@@ -46,10 +61,80 @@ class PostHogWidgetState extends State<PostHogWidget> {
...
+      if (_disposed || !episodeStillCurrent(episode, occluded: true)) {
+        return;
+      }
+      if (accepted) {
+        printIfDebug('Native occlusion started: bridged to native capture.');
+        return;
+      }
+    }
+    printIfDebug('Native occlusion started: emitting placeholder.');
+    final imageInfo = await _screenshotCapturer?.buildOcclusionPlaceholder();
```
ioannisj, line 127: I think buildOcclusionPlaceholder needs a catch. it wraps picture.toImage in try/finally, so a throw there propagates straight out through the uncaught await at line 130, whereas captureScreenshot funnels everything through catchError and resolves to null. can we mirror that here (wrap the body or this call in _onNativeOcclusionChanged, log via printIfDebug, return null) so a failed placeholder just degrades to no frame instead of throwing into the host app?

posthog_flutter/darwin/posthog_flutter/Sources/posthog_flutter/PosthogFlutterPlugin.swift
```diff
@@ -171,10 +190,23 @@ public class PosthogFlutterPlugin: NSObject, FlutterPlugin {
...
+                    sessionReplayConfigMap["captureNativeScreens"] as? Bool ?? false
+                // Unconditional: only bridged captures read these, so a runtime
+                // bridge toggle honors them.
+                if let maskAllTexts = sessionReplayConfigMap["maskAllTexts"] as? Bool {
+                    config.sessionReplayConfig.maskAllTextInputs = maskAllTexts
+                }
+                if let maskAllImages = sessionReplayConfigMap["maskAllImages"] as? Bool {
+                    config.sessionReplayConfig.maskAllImages = maskAllImages
+                }
+                if config.sessionReplay, captureNativeScreens {
```
ioannisj, line 207: small one: when captureNativeScreens is off on a later setup() call the occlusion detector never stops. this start branch (if config.sessionReplay, captureNativeScreens { startOcclusionDetector() }) has no else, so the 1 Hz Timer and the two window observers keep running if someone flips the flag off and re-runs setup. worth an else { stopOcclusionDetector() }? probably rare but cheap to close.

posthog_flutter/lib/src/posthog_widget.dart
turnipdabeets: The `checkApiDart` half should already be green — b8fe251 regenerated the dump yesterday, probably crossed with your review. On the members: agreed on `episodeStillCurrent` — it is actually production logic (three internal call sites), so it moved onto the non-exported `PostHogInternalEvents` (`@internal`, lib/src) and is out of the dump. `debugFlutterCaptureSuppressed` is the awkward one: it reads private instance state, and Dart gives no way to hand that to tests without it being technically public (`hide` only works on top-level symbols, and an extension in another file cannot reach a library-private field). The alternatives are keeping it `@visibleForTesting` or rewriting those five assertions behaviorally (observe that no frame is emitted while suppressed), which is noticeably more test machinery. Kept the annotated getter for now — happy to do the behavioral rewrite if you would rather not ship it at all. c07556c

posthog_flutter/lib/src/posthog_flutter_io.dart
turnipdabeets: Good catch — the real handler had no coverage at all (the episode suite drives the fake). Added three cases to `posthog_flutter_io_test.dart` pushing real calls through `TestDefaultBinaryMessengerBinding`: the field mapping, the missing-argument defaults (episode → 0), and that an unchanged payload still bumps the counter (the monotonic-counter contract the episode logic relies on). c07556c

posthog_flutter/lib/src/posthog_widget.dart
turnipdabeets: Confirmed — a `picture.toImage`/`toByteData` throw propagated straight through the un-guarded await, unlike `captureScreenshot`'s `catchError` funnel. `buildOcclusionPlaceholder` now wraps its body, logs via `printIfDebug`, and resolves to null, so a failed placeholder degrades to no frame (the inner `finally` blocks still handle disposal). c07556c

posthog_flutter/darwin/posthog_flutter/Sources/posthog_flutter/PosthogFlutterPlugin.swift
turnipdabeets: Done — and it needed slightly more than the bare else: a stop mid-episode would leave the Dart side suppressed forever, since nothing would push the final `occluded=false`. The else now calls `disableOcclusionDetector()`, which stops the timer/observers and ends any active episode with an `occluded=false` push. Android had the identical gap, fixed the same way (including a `sessionReplay=false` re-run). c07556c

ioannisj (approved):

PR #474, opened 2026-07-14
feat: add bootstrap option to PostHogConfig

ioannisj:

`bootstrap` seeds identity and feature-flag state at setup, so the first events carry a known distinct ID and `getFeatureFlag()` returns real values before the first `/flags` response. It already ships in `posthog-js`, and it's now the cross-SDK contract in sdk-specs.

posthog-flutter is a wrapper, so this is a pass-through. It adds a `bootstrap` option to `PostHogConfig` and forwards it to the native iOS and Android SDKs, which own all the behavior: identity precedence, flag base-layering, and the first-session drop on `reset()`. I left `sessionID` out on purpose, since the spec scopes it to SDKs with a client-side session id and the native mobile bootstrap config doesn't expose one. On Flutter web `setup` is a no-op, so a web app sets `bootstrap` in its own `posthog.init` call instead.

What's here:

- Dart `PostHogBootstrapConfig` (`distinctId`, `isIdentifiedId`, `featureFlags`, `featureFlagPayloads`) and a `bootstrap` field on `PostHogConfig`, serialized over the method channel.
- The iOS (Swift) and Android (Kotlin) plugins rebuild the native `PostHogBootstrapConfig` and set `config.bootstrap`.
- Native SDK floors point at the released versions that ship bootstrap: posthog-ios `>= 3.66.0` (PostHog/posthog-ios#715), posthog-android `[3.55.0,4.0.0)` (PostHog/posthog-android#618). The android floor matters: 3.54.x still pinned core 6.24.0, which predates `PostHogBootstrapConfig`.
- The compliance adapter's `/init` forwards a `bootstrap` block for the acceptance harness.
- Gitignored local-dev overrides so the example can build against unreleased native branches (see `CONTRIBUTING.md`).

- Dart unit tests for the config serialization: `toMap` present when set and absent when not, plus the identified, anonymous-default, and flags-with-payloads cases.
- Kotlin unit tests for the Android channel decode (`bootstrapConfigFromMap`): full map, empty-map defaults, wrong-type fallbacks. Ran via `./gradlew :posthog_flutter:testDebugUnitTest`.
- Built the example against the released native SDKs from Maven Central and CocoaPods with no local overrides, so the floors resolve the way they will for real consumers.
- CI is green on this head, including `build-android`, `build-apple`, and the SDK compliance tests.

The bootstrap behavior itself (identity precedence, flag layering, reset) is owned and tested by the native SDKs; I haven't done an on-device end-to-end run here.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj

### DIFF
--- a/posthog_flutter/darwin/posthog_flutter.podspec
+++ b/posthog_flutter/darwin/posthog_flutter.podspec
@@ -22,7 +22,7 @@ Postog flutter plugin
   s.osx.dependency 'FlutterMacOS'
 
   # ~> Version 3.61.0 up to, but not including, 4.0.0
-  s.dependency 'PostHog', '>= 3.65.0', '< 4.0.0'
+  s.dependency 'PostHog', '>= 3.66.0', '< 4.0.0'
 
   s.ios.deployment_target = '13.0'
   # PH iOS SDK 3.0.0 requires >= 10.15
--- a/posthog_flutter/darwin/posthog_flutter.podspec
+++ b/posthog_flutter/darwin/posthog_flutter.podspec
@@ -21,7 +21,7 @@ Postog flutter plugin
   s.ios.dependency 'Flutter'
   s.osx.dependency 'FlutterMacOS'
 
-  # ~> Version 3.61.0 up to, but not including, 4.0.0
+  # ~> Version 3.66.0 up to, but not including, 4.0.0
   s.dependency 'PostHog', '>= 3.66.0', '< 4.0.0'
 
   s.ios.deployment_target = '13.0'
--- a/posthog_flutter/darwin/posthog_flutter.podspec
+++ b/posthog_flutter/darwin/posthog_flutter.podspec
@@ -22,7 +22,7 @@ Postog flutter plugin
   s.osx.dependency 'FlutterMacOS'
 
   # ~> Version 3.61.0 up to, but not including, 4.0.0
-  s.dependency 'PostHog', '>= 3.61.0', '< 4.0.0'
+  s.dependency 'PostHog', '>= 3.65.0', '< 4.0.0'
 
   s.ios.deployment_target = '13.0'
   # PH iOS SDK 3.0.0 requires >= 10.15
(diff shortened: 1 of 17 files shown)


posthog_flutter/darwin/posthog_flutter.podspec
```diff
@@ -22,7 +22,7 @@ Postog flutter plugin
   s.osx.dependency 'FlutterMacOS'
 
   # ~> Version 3.61.0 up to, but not including, 4.0.0
```
turnipdabeets: [suggestion] Since we're touching this line — the comment still says 3.61.0 but the floor is now 3.66.0.
```suggestion
  # ~> Version 3.66.0 up to, but not including, 4.0.0
```

posthog_flutter/lib/src/posthog_config.dart
```diff
@@ -303,6 +321,91 @@ class PostHogConfig {
...
+
+  /// Feature flag values served until the first `/flags` response arrives,
+  /// keyed by flag key. Each value is a `bool` for boolean flags or a `String`
+  /// for multivariate flags. Only enabled values are served: `false` or an
+  /// empty string is dropped.
+  final Map<String, Object>? featureFlags;
+
+  /// JSON payloads paired with [featureFlags], keyed by flag key. Each value is
+  /// the already-decoded payload (map, list, string, number, ...).
+  final Map<String, Object>? featureFlagPayloads;
```
turnipdabeets: [question] `featureFlagPayloads` is `Map<String, Object>?` (non-null values), but the native side accepts nullable — Android reads `Map<String, Any?>`. A flag whose JSON payload is literally `null` can't be expressed from Dart here. Intentional, or should this be `Map<String, Object?>`?

turnipdabeets (approved):

posthog_flutter/darwin/posthog_flutter.podspec
ioannisj: Good catch, updated to 3.66.0.

posthog_flutter/lib/src/posthog_config.dart
ioannisj: Not intentional, widened it to `Map<String, Object?>`. posthog-js types this as `Record<string, JsonType>` and `JsonType` includes `null`, so this now matches.

One detail I ran into: posthog-js actually drops falsy bootstrap payloads at runtime (`packages/browser/src/posthog-featureflags.ts:331` filters with `if (bootstrapPayloads[key])`), so a literal `null` payload never gets served there. The type surface matches now though, and the native SDKs decide what they serve.

PR #475, opened 2026-07-15
feat(flags): add $feature_flag_has_experiment to $feature_flag_called events

haacked:

Part of a cross-SDK effort to shrink `$feature_flag_called` events. This first phase adds a `$feature_flag_has_experiment` boolean property to every `$feature_flag_called` event, reflecting the server's `has_experiment` signal: `metadata.has_experiment` in the `/flags?v=2` response and `has_experiment` on `/api/feature_flag/local_evaluation` flag definitions. When the server does not report the field (older deployments, bootstrapped flags), the property is omitted, so it is tri-state: `true`, `false`, or absent (unknown).

This lets ingestion distinguish experiment-linked flag events (which need the full property set for exposure analysis) from the rest, and lets us measure the split before a later phase strips the expensive properties (`$feature/<key>`, `$feature_flag_payload`, per-event system metadata) from non-experiment flag events. This PR intentionally minimizes nothing: no properties are removed, no config options are added, and dedupe behavior is unchanged.

## Changes

The shipped Flutter SDK never constructs `$feature_flag_called` in Dart: iOS/Android delegate to the native SDKs via the method channel and web delegates to posthog-js, all of which are gaining this property in their own PRs. The one place this repo builds the event is the SDK compliance adapter, updated here:

- `sdk_compliance_adapter/lib/adapter_server.dart`: `evaluateFeatureFlag` reads `metadata.has_experiment` from the v2 `flags[key]` entry via a `_flagHasExperiment` helper (omitted when the response carries no boolean `has_experiment`, including legacy `featureFlags`-only responses) and adds the property to the queued event.

No changeset: only the unpublished compliance adapter changed, matching the precedent of prior adapter-only PRs.

New `feature_flag_has_experiment_test.dart` (mock server serving `/flags/?v=2`, recording `/batch`): asserts the property is `true`/`false` with v2 metadata and absent without. `dart format`, `dart analyze`, and `flutter test` on the adapter tests all pass, run inside the repo's compliance image (`ghcr.io/cirruslabs/flutter:3.44.0`). The published `posthog_flutter` package is unmodified.

## :pencil: Checklist

### If releasing new changes

- [ ] Ran `pnpm changeset` to generate a changeset file (not needed; adapter-only change, nothing published)

### DIFF
--- a/sdk_compliance_adapter/lib/adapter_server.dart
+++ b/sdk_compliance_adapter/lib/adapter_server.dart
@@ -494,7 +494,7 @@ class _CompliancePlatform extends PosthogFlutterPlatformInterface {
 
     final client = HttpClient();
     Object? value = false;
-    var hasExperiment = false;
+    bool? hasExperiment;
     try {
       final uri = Uri.parse(_host!).resolve('/flags/?v=2');
       for (var attempt = 0; attempt <= _maxRetries; attempt++) {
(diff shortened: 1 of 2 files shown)


turnipdabeets (approved):
LGTM! Correctly scoped to the compliance adapter (apps get this via the native SDK bumps), and `_flagHasExperiment` is guarded at every level so there's no crash path on legacy or malformed responses.

PR #494, opened 2026-07-17
fix(android): encode replay screenshots off the main thread

turnipdabeets:
## Problem

While session replay records, every snapshot frame runs `BitmapFactory.decodeByteArray` (full screen-size PNG decode), a JPEG re-encode (quality 30), and Base64 **synchronously on the platform main thread** inside the method-channel handler — roughly 25–75ms per frame on midrange hardware, up to once per second for the duration of every recorded session. That is a rhythmic 2–4 dropped-frame hitch, most visible mid-scroll.

iOS already routes the identical work through a background serial queue (`PosthogFlutterPlugin.swift` even carries the comment: *"otherwise we are doing slow operations in the main thread"*) — this brings Android to parity.

## Fix

`sendMetaEvent` and `sendFullSnapshot` are submitted to a dedicated single-threaded executor:

- **Serial by design** so a meta event can never be overtaken by the frame it describes (the ordering contract the Dart side relies on).
- Submissions after engine detach hit the shut-down executor and are dropped with a warning rather than thrown.
- Shut-down executors are recreated on engine reattach. No API or behavior change otherwise; timestamps still come from the SDK clock, captured at enqueue.

## Evidence

Choreographer main-thread gap probe (>17ms) during an identical 20-swipe scripted scroll with replay recording, emulator (Pixel 9, Apple-silicon host — best-case CPU, so deltas understate real devices):

| | stalls >17ms | total blocked | snapshot sends |
|---|---|---|---|
| before | 4 (max 33ms) | 132ms | 49 |
| after | 2 (max 33ms) | 66ms | 47 |

On midrange ARM devices the per-frame cost is several times larger, so the removed stall is proportionally bigger there.

Verified: ktlint clean, Android unit tests pass, replay frames flow end-to-end after the change (sends unchanged).

### DIFF
--- a/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
+++ b/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
@@ -126,9 +126,8 @@ class PosthogFlutterPlugin :
     private val mainHandler = Handler(Looper.getMainLooper())
     private var bitmapExportExecutor = Executors.newSingleThreadExecutor()
 
-    // Frame sends decode and re-encode the image — too slow for the main
-    // thread. Single-threaded so a meta event can never be overtaken by the
-    // frame it describes.
+    // Single-threaded so a meta event can never be overtaken by the frame
+    // it describes.
     private var snapshotSendExecutor = Executors.newSingleThreadExecutor()
 
     // The native SDK stamps its replay events (touches, bridged frames) with
@@ -445,11 +444,9 @@ class PosthogFlutterPlugin :
         }
     }
 
-    // The reply is completed after the worker finishes so Dart's await spans
-    // the encode: a throttleDelay below the encode time then self-regulates
-    // instead of growing the queue without bound. Rejection only happens after
-    // engine detach; the frame is not recoverable then, so it is dropped
-    // rather than thrown.
+    // Replying after the work keeps Dart's await as backpressure: a
+    // throttleDelay below the encode time self-regulates instead of growing
+    // the queue. Rejection means the engine detached; drop, don't throw.
     private fun submitSnapshotWork(
         result: Result,
         work: () -> Unit,
--- a/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
+++ b/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
@@ -126,6 +126,12 @@ class PosthogFlutterPlugin :
     private val mainHandler = Handler(Looper.getMainLooper())
     private val bitmapExportExecutor = Executors.newSingleThreadExecutor()
 
+    // Replay frames are decoded and re-encoded before sending — 25-75ms on
+    // midrange hardware, too slow for the main thread (iOS moved the same
+    // work off-main for the same reason). Single-threaded so a meta event
+    // can never be overtaken by the frame it describes.
+    private val snapshotSendExecutor = Executors.newSingleThreadExecutor()
+
     // The native SDK stamps its replay events (touches, bridged frames) with
     // config.dateProvider, which prefers the network-time clock on API 33+ and
     // can diverge from System.currentTimeMillis. Replay is ordered by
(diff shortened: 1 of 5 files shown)


posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -970,7 +992,7 @@ class PosthogFlutterPlugin :
             val x = call.argument<Int>("x") ?: 0
             val y = call.argument<Int>("y") ?: 0
             if (imageBytes != null) {
-                snapshotSender.sendFullSnapshot(imageBytes, id, x, y)
+                submitSnapshotWork { snapshotSender.sendFullSnapshot(imageBytes, id, x, y) }
```
dustinbyrne: [P2] Consider capturing the replay timestamp before submitting the encoding work

`SnapshotSender` currently reads the SDK clock after bitmap decoding and JPEG/Base64 encoding. Flutter iOS instead captures `Date` before dispatching to its serial queue.

At the default capture interval this should only introduce tens of milliseconds of skew, so I don’t think it needs to block this improvement. Capturing the SDK-backed timestamp here and passing it into `SnapshotSender` would nevertheless provide closer iOS parity and prevent larger skew if the executor ever becomes backlogged.

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -681,6 +702,7 @@ class PosthogFlutterPlugin :
         stopOcclusionDetector()
         channel.setMethodCallHandler(null)
         bitmapExportExecutor.shutdown()
+        snapshotSendExecutor.shutdown()
```
dustinbyrne, line 729: [P2] Reattaching the same plugin instance leaves snapshot delivery disabled

`snapshotSendExecutor` is created once, shut down here, and not recreated in `onAttachedToEngine`. If Flutter removes and later re-adds the same plugin instance, both snapshot handlers report success while their work is rejected by the terminated executor.

Generated registration normally creates a fresh instance, so this is an uncommon lifecycle edge. Consider recreating attachment-scoped executors on attach, with coverage for attach → send → detach → reattach → send.

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -126,6 +126,11 @@ class PosthogFlutterPlugin :
     private val mainHandler = Handler(Looper.getMainLooper())
     private val bitmapExportExecutor = Executors.newSingleThreadExecutor()
 
+    // Frame sends decode and re-encode the image — too slow for the main
+    // thread. Single-threaded so a meta event can never be overtaken by the
+    // frame it describes.
+    private val snapshotSendExecutor = Executors.newSingleThreadExecutor()
```
dustinbyrne: [P2/design] Consider whether the snapshot queue needs backpressure

This executor has an unbounded queue, while the method-channel call now succeeds immediately after submission. With the default one-second capture interval the worker has ample headroom, but a very low custom `throttleDelay` could enqueue full-screen image buffers faster than they are encoded.

Consider completing the channel result after worker completion, or bounding/coalescing queued frames, if very low capture intervals are intended to be supported safely.

dustinbyrne (approved):
The core performance diagnosis is corroborated: Android currently performs bitmap decode, JPEG re-encoding, and Base64 work on the platform thread, and this change moves it to a serial worker while preserving meta → frame ordering. The remaining comments are non-blocking follow-ups.

Human-driven, agent-assisted review.

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
turnipdabeets: Agreed — our pre-PR pass flagged the same thing. Fixed in a11ae47: the handler captures the SDK-clock timestamp at enqueue and passes it into SnapshotSender, matching the iOS shape.

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
turnipdabeets: Fair — generated registration makes this unreachable today, but it's cheap insurance. Fixed in a11ae47: both attachment-scoped executors are recreated on attach, with the attach → send → detach → reattach → send coverage you suggested.

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
turnipdabeets: Good catch — with the result completing at enqueue, a throttleDelay below the encode time does grow the queue without bound. Fixed in a11ae47 with your first option: the reply completes after the worker finishes (posted via the main handler), restoring the Dart-side one-in-flight backpressure while keeping the encode off the main thread. E2E-verified frames still flow at the throttle cadence.

PR #499, opened 2026-07-24
feat(replay): mask canvas session replay recordings on Flutter web

### LINKED ISSUE #496: Flutter web canvas recordings don't support masking of painted Text widgets
### Bug Description

Title: Flutter web canvas recordings don't support masking of painted Text widgets

Description:

When using Flutter web with CanvasKit rendering and Canvas capture enabled for session replay, there appears to be no supported way to mask Text widgets that are painted directly to the canvas.

Environment:

Platform: Flutter web (CanvasKit)
PostHog Flutter SDK: 5.11.0
Canvas capture: Enabled in project settings
Widget types: Plain Text widgets displaying sensitive PII (user names, emails)
Expected behavior:
Setting config.sessionReplayConfig.maskAllTexts = true should mask text content in session replays, including text rendered to canvas on Flutter web.

Actual behavior:

maskAllTexts and maskAllImages in sessionReplayConfig appear to only apply to mobile screenshot-based replay (iOS/Android)
On Flutter web, these config options are no-ops during setup()
JavaScript web SDK masking options like maskTextSelector can't reach inside canvas pixels
Sensitive text rendered by Flutter to the canvas remains visible in session replays
Steps to reproduce:

Set up Flutter web app with CanvasKit renderer
Initialize PostHog with session replay enabled:
dart
final config = PostHogConfig('<ph_project_token>');
config.sessionReplay = true;
config.sessionReplayConfig.maskAllTexts = true;
await Posthog().setup(config);
Enable Canvas capture in PostHog project settings
Display sensitive data using Text widgets
Record and view session replay
Observe that text is still visible
Questions:

Is masking of canvas-rendered text supported for Flutter web?
If not, are there plans to add this capability?
Should the documentation clarify that sessionReplayConfig masking only applies to mobile platforms?
Additional context:
The [Flutter session replay installation docs](https://posthog.com/docs/session-replay/installation/flutter) and [privacy controls docs](https://posthog.com/docs/session-replay/privacy) don't explicitly distinguish between mobile and web behavior for these masking options.

### Debug info

```shell
- [ ] PostHog Cloud, Debug information: [please copy/paste from https://us.posthog.com/settings/project-details#variables or https://eu.posthog.com/settings/project-details#variables]
- [ ] PostHog Hobby self-hosted with `docker compose`, version/commit: [please provide]
- [ ] PostHog self-hosted with Kubernetes (deprecated, see [`Sunsetting Kubernetes support`](https://posthog.com/blog/sunsetting-helm-support-posthog)), version/commit: [please provide]
```

turnipdabeets:

Fixes #496 — on Flutter web, session replay is posthog-js canvas capture, and the entire app (including PII text) is pixels inside the CanvasKit canvas. DOM-based masking can't see it, so `sessionReplayConfig.maskAllTexts`/`maskAllImages` **and `PostHogMaskWidget`** were all silent no-ops on web.

> **Depends on PostHog/posthog-js#4270 — shipped in posthog-js 1.408.0**, and `_minPosthogJsVersion` is pinned to `1.408.0`. On older posthog-js the plugin still registers and never throws (verified live against released 1.407.5: the `flt-semantics-host` accessibility exclusion works there, `set_config` with a function is accepted, the restart is harmless) — but canvas frames are recorded **unmasked**, so the plugin emits a one-time console warning when the detected `posthog.version` is confirmed older than the minimum (absent/unparseable versions assume-new, so the gate can never misfire on future builds).

### How you turn it on

Web replay is configured in `posthog.init`, so the opt-in lives there too — declare the mask-provider slot and the plugin fills it once Flutter has booted:

```js
posthog.init('<token>', {
  session_recording: {
    captureCanvas: { recordCanvas: true },
    canvasCapture: {
      // plugin replaces this once Flutter starts; until then frames are
      // skipped rather than recorded unmasked
      maskRegionsFn: () => null,
    },
  },
})
```

**Since #501 was merged into this branch, mounting a `PostHogMaskWidget` is a second, self-contained opt-in** — the most explicit possible statement of "never record this" should not silently no-op just because `web/index.html` wasn't edited. The first mount registers the provider itself (idempotent: one `set_config`, one restart, however many mask widgets mount, whichever side loads first). Prefer the `posthog.init` declaration when you can: it covers the frames captured before Flutter boots (skipped instead of recorded unmasked) and avoids the one-time mid-session recording restart the mount path needs. One mount enables the whole masking configuration (`maskAllTexts`/`maskAllImages` default true), matching iOS/Android semantics.

**With neither the key nor a mask widget, the plugin registers nothing** — no `set_config`, no recording restart, no `blockSelector` change — and recording behaves exactly as posthog-js is configured, as it does today. The mobile-only `config.sessionReplay` flag has **no effect on web** (it gates the native iOS/Android recorder; overloading it here would have made a flag documented as "Android and iOS" silently govern web behavior).

### What it does

- **Rect computation reuses the mobile masking logic** — one widget-tree walk (`PostHogMaskController.getMaskElements`, new single-walk method) with the exact selection rules mobile uses: `maskAllTexts`/`maskAllImages` gate the full text/image sets; `PostHogMaskWidget` and obscured text fields always mask. Rects are transform-aware (AABB for rotated/scaled widgets), outset 1px, converted to canvas-relative CSS px against the `flutter-view` host rect (correct for embedded/offset views), and cached per Flutter frame (idle cost ≈0.01ms/call, measured).
- **posthog-js paints the regions black inside its capture pipeline** before frames are encoded — masked pixels never leave the browser. The live canvas is untouched; masking happens on a copy in the encode worker.
- **Fail-closed**: a failed widget-tree walk returns `null`, which makes posthog-js **skip that frame** rather than ship it unmasked (one-time console warning after ~10 consecutive failures explains the `PostHogWidget` requirement). Declaring `maskRegionsFn: () => null` extends the same protection to frames captured before Flutter boots. Non-Flutter canvases get an empty region list (identity-checked via the flutter-view host chain), so they are recorded normally; a canvas belonging to a *different* Flutter view on a multi-view page is skipped (`null`) instead — our rects must never ship with someone else's view.
- **Accessibility side-channel closed**: with a11y active, Flutter mirrors widget text into `flt-semantics` DOM nodes as plain text, which rrweb records; the provider forwards `blockSelector: 'flt-semantics-host'` (merged with any user selector). Reproduced and verified gone.
- **Lifecycle**: posthog-js reads `blockSelector` only at rrweb `record()` start, so an in-flight recording is restarted once after registration (same session id; sampling decision persists). Registration retries with backoff (250ms→4s), then keeps polling at the 4s cap indefinitely. A `window.posthog` that exists but hasn't finished `init` (posthog-js's `__loaded` flag false — e.g. consent-gated init) counts as not-ready and keeps being retried; only an initialized config without the `maskRegionsFn` key is treated as not opted in. Covers the snippet stub being replaced by the real instance, posthog-js loading after Flutter entirely, and `posthog.init` running arbitrarily late. (`__loaded` is undocumented but de-facto-stable posthog-js API: set when `_init` completes, absent/false on the stub and on a constructed-but-uninitialized instance.)

- **Mount opt-in machinery (from #501)**: the apply path is a small state machine (`applied` / `notOptedIn` / `posthogNotReady`) so "posthog-js is up but the app didn't opt in" parks instead of terminating — a later mask-widget mount re-enters it. A mask widget outside the tracked `PostHogWidget` tree cannot fail open: every `maskRegionsFn` call revalidates all mounted mask widgets against the tracked tree and skips the frame if any sits outside it. The tracked-tree boundary resolves the same route-dependent way the masking walk does (root navigator when `PostHogWidget` sits under an active route), so a mask widget in a root-navigator dialog is covered, not rejected. Registration deliberately doesn't wait on `recordCanvas` — remote config can enable it after init, and the `blockSelector` accessibility exclusion is worth applying even with canvas capture off.

No new public Dart API (package barrel diff is zero lines). Existing config options and `PostHogMaskWidget` start working on web with mobile-identical semantics, once the provider is declared or a `PostHogMaskWidget` mounts.

- `flutter analyze` clean; full suite 247 passing (VM run also proves the conditional-import io stubs compile on mobile); the two web test files run 50 chrome-platform tests in CI. VM tests for the new code (geometry incl. rotation AABB + degenerate rects; single-walk equivalence vs the two-walk union); browser tests 20/20 (`--platform chrome`, wired into CI): set_config merge preserving user config, **no registration at all when the provider isn't declared** (no `set_config`, no restart), the canvas-relative coordinate conversion pinned to exact pixels, fail-closed `null` for a flutter-view canvas without PostHogWidget, warn-once on persistent walk failure, foreign-canvas exclusion, semantics blockSelector with maskAllTexts=false, deferred registration + stub-replacement convergence (incl. posthog-js absent entirely at setup), restart-exactly-once, posthog-js present-but-uninitialized (`__loaded` false) applying once init declares the provider, and a throwing retry tick rescheduling instead of killing the chain, plus the version-gate warning matrix (older posthog-js warns once but still registers; newer/absent/garbage versions never warn). The #501 additions pin the mount path: mask-widget mounts opting in with nothing declared (before or after posthog-js appears), exactly one `set_config`/restart across both opt-in paths and repeated mounts, outside-tree mask widgets forcing frames to `null` per-frame (and recovering on removal), the root-navigator-dialog shape producing regions, the multi-view ambiguous-host fail-closed, and a failed mount-triggered apply retrying instead of consuming the opt-in.
- End-to-end against the posthog-js branch from PostHog/posthog-js#4270: wire-level canvas frames decoded at `before_send` and inspected pixel-by-pixel across four configurations — no provider declared (nothing masked), `maskAllTexts`+`maskAllImages` (everything masked), text-only (photos correctly left visible), and neither flag (only `PostHogMaskWidget` subtrees masked). Rotated and scaled text verified to mask via their transform-aware AABBs. Recordings played back in the production PostHog player.
- Perf: cold walk ~20µs per masked text node (200 widgets ≈ 4.4ms, 1000 ≈ 20.5ms, chrome/debug); per-frame cache makes steady-state ≈0.01ms. O(N) walk optimization (shared with mobile) is a measured follow-up.

Known limitations (documented here and in the dartdoc; the changeset is a one-line entry per the changelog style, and a posthog.com docs PR will carry the full setup guidance):
- `CustomPainter`-drawn text isn't in the widget tree → not masked (same as mobile; `PostHogMaskWidget` is the escape hatch).
- **Mount-path caveats** (when relying on `PostHogMaskWidget` alone, without the `posthog.init` declaration): frames captured before the first mount are recorded unmasked, and the recording restarts once when masking switches on (the replay shows a split there) — `blockSelector` is a start-time option in posthog-js. Documented on the widget's dartdoc.
- **A client-side `blockSelector` takes precedence over the project-level one.** The plugin adds `flt-semantics-host` via `set_config`, and posthog-js resolves `client ?? server`, so a selector configured only in project settings is superseded. The plugin cannot merge with it — the server value is not exposed on `posthog.config`. Will be called out in the docs PR.
- **Platform views are DOM, not canvas pixels.** `HtmlElementView`-based widgets (maps, webviews) are recorded by posthog-js's DOM rules; `PostHogMaskWidget` around one does not mask it on web.
- Requires `PostHogWidget` wrapping the app — without it, masking fails closed (frames skipped) by design.

## :pencil: Checklist

- [x] I updated the docs if needed. (Changeset added; posthog.com docs PR drafted, to land with the releases.)

### If releasing new changes

- [x] Changesets added (`.changeset/canvas-masking-web.md` — the feature; `.changeset/mask-text-flag-fix.md` — the cross-platform `maskAllTexts: false` fix), both one-line entries per the changelog style.

## 🤖 Agent context

Built with Claude Code from the investigation of #496 (transferred from PostHog/posthog#66291). The provider deliberately reuses the mobile masking walk rather than a parallel implementation, so web and mobile masking semantics cannot drift.

Two design changes during review, both narrowing the public surface: posthog-js dropped its `requireMaskProvider` boolean in favour of the provider declaration itself carrying the opt-in, and this PR dropped an earlier `config.sessionReplay` gate for the same reason — it would have given a flag documented as iOS/Android-only a silent web meaning, and its `false` default would have made unmasked recording the default for anyone enabling canvas capture.

Notable findings fixed during pre-PR review loops (several with executed repros): posthog-js replaces the snippet stub rather than upgrading it (retry must re-read `window.posthog`), `set_config` against the stub would wipe user `session_recording` config, start-time-only options require the one-time recording restart, and the a11y semantics tree leaks widget text as DOM text nodes.

Rebased on #500 (`fix(replay): mask every element that matched a masking rule`), which fixed a shared masking bug this work surfaced — `extractRects()` dropped `PostHogMaskWidget` wrappers with more than one masked child, affecting shipped iOS/Android as well as web.


posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,396 @@
...
+  List<Rect>? _currentContainerRects() {
+    if (_cachedContainerRects != null && _cachedAtFrame == _frameCount) {
+      return _cachedContainerRects;
+    }
+
+    final replayConfig = _config.sessionReplayConfig;
+    final elements = PostHogMaskController.instance.getMaskElements(
+      includeAllWidgets:
+          replayConfig.maskAllTexts || replayConfig.maskAllImages,
+    );
```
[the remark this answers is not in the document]
turnipdabeets: The walk cannot produce text elements when `maskAllTexts` is false: parsing is gated at parser registration, not in this call — `element_parsers_const.dart` only registers the `RenderParagraph`/`RenderEditable`/`RenderTransform` parsers when `maskAllTexts` is true, and `element_object_parser.dart` only relates render objects that have a registered parser. `includeAllWidgets` just switches between wrapper-only and full-walk extraction over that already-config-gated tree, so images-on/text-off masks images only. We verified the per-type gating live in the mirrored direction (text-on/images-off leaves photos visible on the wire).

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,396 @@
...
+
+  // shared so a second provider's cache is not compared against a counter that
+  // never advances; the callback cannot be removed once added
+  static int _frameCount = 0;
+  static bool _frameCallbackRegistered = false;
+
+  void register() {
+    try {
+      _active = this;
+      _registerUnsafe();
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — fixed in 207f4c1: `register()` now cancels the predecessor's pending retry timer before taking over, with a regression test (two registers, then the stub appears → exactly one `set_config` and one restart; we confirmed the test fails without the cancel).

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,396 @@
...
+    required double width,
+    required double height,
+  });
+}
+
+@JS('Object.assign')
+external JSObject _objectAssign(JSObject target, JSObject source);
+
+const _semanticsBlockSelector = 'flt-semantics-host';
+
```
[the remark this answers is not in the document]
turnipdabeets: Intentional: the `0.0.0` placeholder keeps the gate inert because the qualifying posthog-js release does not exist yet — PostHog/posthog-js#4270 is unreleased and this PR stays gated on it. Pinning the constant to that release number is a tracked release-blocking step in the PR's dependency note, the changeset, and the comment above the constant.

.changeset/canvas-masking-web.md
```diff
@@ -0,0 +1,70 @@
...
+- Widgets rendered as DOM rather than canvas pixels — `HtmlElementView`-based platform
+  views such as maps, webviews and iframes — are recorded as DOM and masked by
+  posthog-js's DOM rules, not by canvas mask regions. `PostHogMaskWidget` around a
+  platform view does not mask it on web.
+- Your app must be wrapped in `PostHogWidget`. If it is not, canvas frames are
+  skipped instead of recorded unmasked, and a console warning explains the fix.
+- Requires a posthog-js version that supports `canvasCapture.maskRegionsFn`
+  (minimum version pinned at release). On an older posthog-js the plugin still
+  registers — the accessibility-tree exclusion works there — but canvas frames
+  are NOT masked, and a console warning tells you to upgrade.
```
marandaneto: this is a huge wall of text, i'd leave out and add to the docs if needed

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,402 @@
...
+    }
+    final canvasRect = canvas.getBoundingClientRect();
+    final hostRect = host.getBoundingClientRect();
+    final offset = containerOrigin +
+        Offset(hostRect.left, hostRect.top) -
+        Offset(canvasRect.left, canvasRect.top);
+
+    final regions = <JSObject>[];
+    for (final rect in containerRects) {
+      final shifted = rect.shift(offset);
```
marandaneto: Could we apply the container's full transform here rather than only shifting by `localToGlobal(Offset.zero)`? These rects are container-local, so an ancestor scale, rotation, or skew changes the painted geometry without changing the emitted mask dimensions. I reproduced this with `Transform.scale(scale: 2)`: the global content was twice the emitted mask size, leaving pixels outside the mask.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,402 @@
...
+  }
+
+  // with accessibility enabled, Flutter mirrors widget text into the
+  // flt-semantics DOM tree, which rrweb would otherwise record in plaintext
+  void _blockSemanticsHost(JSObject sessionRecording) {
+    final existing = sessionRecording.getProperty<JSAny?>('blockSelector'.toJS);
+    var selector = _semanticsBlockSelector;
+    if (existing.isA<JSString>()) {
+      final current = (existing as JSString).toDart;
+      if (current.contains(_semanticsBlockSelector)) {
```
marandaneto: `contains` can produce a false positive for an unrelated selector such as `.not-flt-semantics-host`. In that case we return without adding the actual `flt-semantics-host` selector, so rrweb can still serialize Flutter's semantics text. Could we append it unless it is present as an actual selector, or simply append it unconditionally?

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,402 @@
...
+    try {
+      return _unsafeComputeMaskRegions(canvas);
+    } catch (e) {
+      printIfDebug('PostHog: error computing canvas mask regions: $e');
+      return null;
+    }
+  }
+
+  JSArray<JSObject>? _unsafeComputeMaskRegions(web.HTMLCanvasElement canvas) {
+    final host = _flutterViewHost(canvas);
```
marandaneto, line 434: This accepts a canvas under any `flutter-view`, but `_currentContainerRects()` always walks the single global `PostHogMaskController.containerKey`. On a page with multiple Flutter views, a second canvas receives rectangles from the first view and can be recorded unmasked. Could we verify that this host corresponds to the controller's Flutter view and fail closed otherwise?

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,402 @@
...
+
+  // rects only change when Flutter paints a frame, so cache per frame instead
+  // of recomputing on every posthog-js canvas tick
+  List<Rect>? _currentContainerRects() {
+    if (_cachedContainerRects != null && _cachedAtFrame == _frameCount) {
+      return _cachedContainerRects;
+    }
+
+    final replayConfig = _config.sessionReplayConfig;
+    final elements = PostHogMaskController.instance.getMaskElements(
```
marandaneto, line 605: `PostHogMaskController.instance.parsers` is frozen from the config present when the singleton first initializes. A later `setup()` with stronger masking settings creates this provider with the new config, but the walk still uses the old parser map—for example, changing `maskAllImages` from false to true leaves `RenderImage` absent. Could setup refresh the parsers, or could the controller derive selection from the current config?

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,402 @@
...
+      includeAllWidgets:
+          replayConfig.maskAllTexts || replayConfig.maskAllImages,
```
marandaneto, line 607: I reproduced the mixed-flags case in an isolated widget test: with `maskAllTexts = false` and `maskAllImages = true`, ordinary `Text` still appears in `getMaskElements`. `ElementObjectParser` relates `Text` widgets unconditionally, independently of the registered `RenderParagraph`/`RenderEditable` parsers, and this OR enables the full extraction. Could we keep text and image selection independent here?

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,402 @@
...
+
+    final config = JSObject();
+    config.setProperty('session_recording'.toJS, sessionRecording);
+    ph.set_config(config);
+
+    // blockSelector is only read when rrweb's record() starts, so an in-flight
+    // recording must be restarted
+    if (ph.sessionRecordingStarted()) {
+      ph.stopSessionRecording();
+      ph.startSessionRecording();
```
marandaneto, line 310: If `stopSessionRecording()` succeeds but `startSessionRecording()` throws, the retry sees `sessionRecordingStarted()` as false and skips both calls, leaving recording permanently stopped. The current regression test covers `stopSessionRecording()` throwing, but not this partial restart. Could we preserve the restart-needed state across the retry?

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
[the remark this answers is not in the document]
turnipdabeets: Correction: this finding was right and our earlier reply was wrong. The parser-map gating we cited only covers the `RenderImage`/`RenderParagraph`/`RenderEditable` branches — `element_object_parser.dart` also has an `element.widget is Text` branch above them that related Text widgets unconditionally, so `maskAllTexts: false` + `maskAllImages: true` did mask all text (on mobile screenshots too, same shared walk). Fixed in e7d982c by gating that branch on the current `maskAllTexts`, with VM widget tests for both flag directions.

.changeset/canvas-masking-web.md
turnipdabeets: Trimmed in 551ff8c — down to the init snippet, the fail-closed bullets, and one caveats paragraph; the deep mechanics move to the docs PR.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
turnipdabeets: Good catch — fixed in 7b8480d: rects now go through the container's full `getTransformTo(null)` matrix (`MatrixUtils.transformRect`, same AABB approach as element-level transforms) instead of the origin shift. Regression test: `Transform.scale(2)` around `PostHogWidget` doubles the emitted region.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
turnipdabeets: Fixed in 7b8480d — exact token match after splitting on commas; `.not-flt-semantics-host` now gets `flt-semantics-host` appended alongside it (tested).

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
turnipdabeets: Fixed in 7b8480d: we resolve our view's host (`View.maybeOf` → `ui_web.views.getHostElement`) and fail closed (frame skipped) for a flutter-view canvas outside it; an unresolvable host with more than one flutter-view in the document also fails closed. One empirical caveat: in full-page mode `getHostElement` returns `<body>`, which can't discriminate — so full-page keeps today's behavior and the check bites for embedded/multi-view hosts (the foreign-view test simulates embedded hosts via a test-only override).

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
turnipdabeets: Fixed in 7b8480d — added `PostHogMaskController.refreshParsers(config)`, called from the provider's `register()`, so web setup rebuilds the map from the current config. Mobile call sites untouched.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
turnipdabeets: You're right — and it affects shipped mobile screenshots too (same shared walk). Fixed in e7d982c by gating the `is Text` branch on the current `maskAllTexts` (same pattern as the TextField branch below it); VM widget tests cover both flag directions, and the changeset notes the mobile-facing fix.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
turnipdabeets: Fixed in 7b8480d — a pending-restart flag is set before `stop` and cleared only after `start` succeeds, so a retry after a throwing `start()` still calls it even though `sessionRecordingStarted()` reads false. The test models the recording-state transition with `start` throwing once.

.changeset/canvas-masking-web.md
```diff
@@ -0,0 +1,52 @@
...
+Caveats: web replay is configured in `posthog.init` (the Dart `sessionReplay`
+flag drives iOS/Android only); list any project-level `blockSelector` in
+`posthog.init` too, as the client-side selector takes precedence; DOM-rendered
+platform views (`HtmlElementView`) follow posthog-js's DOM masking rules, not
+canvas mask regions; on a page embedding multiple Flutter views, canvases
+belonging to other Flutter views are skipped entirely (not recorded), since this
+plugin's mask regions only describe its own view.
+
+Also fixes `maskAllTexts: false` still masking `Text` widgets when
+`maskAllImages` is on — this applies to iOS/Android screenshot masking too.
```
marandaneto: changelog should be small and concise, add this info to docs if needed and link to it if needed

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,472 @@
...
+    return regions.toJS;
+  }
+
+  // In full-page mode the embedder host is <body>, which contains every
+  // flutter-view on the page, so only a view embedded in a dedicated host
+  // element (multi-view) can be told apart from ours.
+  bool _isOwnViewCanvas(web.Element canvasViewHost) {
+    final ownHost = debugOwnViewHostOverride ?? _resolveOwnViewHost();
+    if (ownHost != null) {
+      return ownHost.contains(canvasViewHost);
```
marandaneto, line 527: blocking: In full-page mode `getHostElement` returns `<body>`, so this is true for every `flutter-view` on the page. If the page also hosts another Flutter view/engine, the global callback supplies that foreign canvas with rectangles from this controller's view instead of returning `null`; mismatched coordinates can leave sensitive pixels recorded unmasked. Could we compare the closest `flutter-view`'s view id with `View.maybeOf(context)?.viewId`, or otherwise fail closed when `<body>` contains more than one Flutter view?

marandaneto (approved):
left a last comment, approving to unblock

.changeset/canvas-masking-web.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog_flutter": minor
```
marandaneto, line 2: this PR has 2 changesets, why?

turnipdabeets: merging #501 into this.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
turnipdabeets: Fixed in 863317e. The precise version isn't reachable with public API — `ui_web.views` (`FlutterViewManagerProxy`) only exposes `getHostElement`/`getInitialData`, so a canvas's closest `flutter-view` can't be mapped to a view id. Went with your fallback instead: when the resolved host element contains more than one `flutter-view`, ownership can't be proven and every canvas's frames are skipped (one-time debug log). A host with a single `flutter-view` keeps the exact containment test, so single-view pages are unchanged. Pinned by a new chrome test.

ioannisj (approved):

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
turnipdabeets: CI is green on the fix (863317e) — resolving. Follow-up hardening also landed on the #501 branch via ReviewHog: the tracked-tree membership checks now share the walk's route-dependent root resolution (e89ff8e), so the two gates and the walk can't drift again.

.changeset/canvas-masking-web.md
turnipdabeets: Intentional: one is the web feature (minor), the other fixes a pre-existing cross-platform bug (`maskAllTexts: false` still masking `Text` when `maskAllImages` is on) that ships to iOS/Android too — separate entries so mobile users see the fix in the changelog rather than it being buried in a web feature line. With #501's entry folded into the feature changeset on merge, the combined PR ships exactly these two.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,472 @@
...
+    }
+    // unresolvable: with a single flutter-view it can only be ours
+    return web.document.querySelectorAll('flutter-view').length <= 1;
+  }
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — fixed in 46364cc: the no-host fallback now claims ownership only when exactly one light-DOM `flutter-view` exists and it contains the canvas's own view host; a zero count (shadow-embedded views) fails closed.

posthog_flutter/lib/src/replay/web/web_canvas_mask_geometry.dart
```diff
@@ -0,0 +1,23 @@
...
+    final transform = element.transform;
+    final rect = (transform != null
+            ? MatrixUtils.transformRect(transform, element.rect)
+            : element.rect)
+        // outset so capture-resolution rounding can't leave a sub-pixel glyph
+        // edge visible at the mask border
+        .inflate(1.0);
+    if (!rect.isFinite || rect.isEmpty) {
+      continue;
+    }
```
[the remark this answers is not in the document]
turnipdabeets: Fixed in 46364cc — the isFinite/isEmpty check now runs on the raw rect before the 1px outset, so genuinely zero-size rects are dropped; pinned in the degenerate-rect geometry test.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,472 @@
...
+          : MatrixUtils.transformRect(containerTransform, rect);
+      final shifted = globalRect.shift(offset);
+      regions.add(_JSMaskRegion(
+        x: shifted.left,
+        y: shifted.top,
+        width: shifted.width,
+        height: shifted.height,
+      ));
+    }
+    return regions.toJS;
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — fixed in 46364cc: any non-finite region after the container transform now returns `null` for the whole frame (fail closed, frame skipped) instead of shipping a silently no-op mask. Pinned by a new singular-transform test.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,472 @@
...
+    }
+    _consecutiveWalkFailures = 0;
+
+    // our rects always describe PostHogWidget's tree — shipping them with a
+    // different flutter-view's canvas would record that view unmasked
+    if (!_isOwnViewCanvas(host)) {
+      return null;
+    }
+    if (containerRects.isEmpty) {
+      return JSArray<JSObject>();
```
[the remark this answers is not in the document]
turnipdabeets: Fixed in 46364cc — ownership and mask-widget revalidation are checked before the walk and its failure accounting, so foreign canvases no longer advance the warn-once counter or cost a walk.

posthog_flutter/lib/src/replay/mask/posthog_mask_widget.dart
```diff
@@ -4,6 +4,11 @@ import 'package:flutter/material.dart';
 ///
 /// Wrap sensitive UI with [PostHogMaskWidget] to hide that area in captured
 /// screenshots, regardless of the global session replay masking settings.
+///
+/// **Flutter web:** this has no effect unless canvas masking is enabled, since
+/// the canvas is masked by posthog-js rather than by this plugin. Declare
+/// `session_recording: { canvasCapture: { maskRegionsFn: () => null } }`
+/// in your `posthog.init` call to turn it on. iOS and Android need no setup.
```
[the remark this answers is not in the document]
turnipdabeets: Fixed in 46364cc — the dartdoc snippet now includes `captureCanvas: { recordCanvas: true }` and mentions the project-settings Canvas capture toggle as the alternative.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,472 @@
...
+    final recordCanvas =
+        (captureCanvas as JSObject).getProperty<JSAny?>('recordCanvas'.toJS);
+    if (recordCanvas.isA<JSBoolean>() && (recordCanvas as JSBoolean).toDart) {
+      web.console.warn(
+        'PostHog: canvas session recording is enabled but maskRegionsFn '
+                'is missing from posthog.init, so text painted by Flutter is recorded '
+                'unmasked. See the posthog_flutter CHANGELOG for the snippet.'
+            .toJS,
+      );
+    }
```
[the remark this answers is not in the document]
turnipdabeets: Superseded by the mount-based opt-in merged from #501: the developer this describes (flags off, relying on `PostHogMaskWidget`, no declaration) is auto-registered the moment their first mask widget mounts — the silent-unmasked shape no longer exists. With flags off and no mask widget mounted there is nothing sensitive marked to warn about; the residual pre-mount window is a documented caveat in the PR body and dartdoc.

posthog_flutter/lib/src/replay/mask/posthog_mask_controller.dart
```diff
@@ -11,24 +11,36 @@ import 'package:posthog_flutter/src/replay/mask/widget_elements_decipher.dart';
...
     ).parsersMap;
   }
 
+  /// Rebuilds the parser map for [config]. The singleton captures the config
+  /// present at first access, which a later `setup()` with different masking
+  /// flags would otherwise never update.
+  void refreshParsers(PostHogSessionReplayConfig? config) {
+    parsers = _buildParsers(config);
+  }
+
```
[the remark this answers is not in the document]
turnipdabeets: Accurate observation, deliberately out of scope: the stale-parser-map singleton predates this PR on native, and wiring `refreshParsers()` into the shared `setup()` changes iOS/Android behavior that this web PR should not carry. We will file a follow-up for the native path; the doc comment describes the mechanism the web call site uses.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -0,0 +1,472 @@
...
+      PostHogMaskController.instance
+          .refreshParsers(_config.sessionReplayConfig);
+      _registerUnsafe();
+    } catch (e) {
+      // a partial first apply (set_config landed, restart threw) must not end
+      // the chain — the retry re-runs the whole apply, which is idempotent
+      printIfDebug('PostHog: failed to register web canvas masking: $e');
+      _scheduleRetry(const Duration(milliseconds: 250));
+    }
+  }
```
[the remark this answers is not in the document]
turnipdabeets: Previously triaged and deferred in an earlier review round: the trigger needs posthog-js `startSessionRecording()` to throw (only ever simulated by throwing stubs) AND a second `Posthog().setup()` inside the ≤250ms retry window, and the proposed static `_pendingRestart` couples restart state across provider instances — a worse trade than the risk. The single-provider recovery path is pinned by the restart-retry tests.

posthog_flutter/lib/src/replay/mask/posthog_mask_controller.dart
```diff
@@ -61,6 +73,39 @@ class PostHogMaskController {
...
+      final widgetElementsTree = _widgetScraper.parseRenderTree(context);
+
+      if (widgetElementsTree == null) {
+        printIfDebug('Error: widgetElementsTree is null after parsing.');
+        return null;
+      }
+
+      return [
+        ...widgetElementsTree.extractMaskWidgetRects(),
+        if (includeAllWidgets) ...widgetElementsTree.extractRects(),
```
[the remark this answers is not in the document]
turnipdabeets: By design — byte-for-byte parity with mobile's two-walk union is an explicit goal of this PR, pinned by the single-walk-equivalence test; the consequence of the overlap is painting black over already-black pixels. Deduplicating web-only would diverge from mobile, and deduplicating the shared walk is a cross-platform change that belongs in its own PR.

posthog_flutter/lib/src/replay/mask/posthog_mask_controller.dart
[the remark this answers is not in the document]
turnipdabeets: Follow-up filed: #507 wires `refreshParsers()` into the shared `Posthog().setup()` so iOS/Android honor changed masking flags too (stacked on this branch).

PR #500, opened 2026-07-27
fix(replay): mask every element that matched a masking rule

turnipdabeets:

`ElementData.extractRects()` — the function that produces the list of rects the replay capturer paints black over each screenshot — was not recursive. It walked the root's direct children, and grandchildren only when a child had more than one child (dropping the child itself in that branch). Anything else was silently discarded.

Two consequences, both in shipped code:

1. **A `PostHogMaskWidget` that wraps more than one masked child loses its own mask rect.** Its matched children (the texts) are still masked, but anything the wrapper covers that does not match a masking rule on its own — an image while `maskAllImages` is off, an avatar, a decoration, the space between children — is recorded in the clear, despite the developer explicitly asking for that subtree to be masked.
2. **Matches nested more than one level deep are dropped**, so masking depended on where in the widget tree an element happened to land.

Point 1 is the one that actually leaks today. It bites specifically when `maskAllTexts` or `maskAllImages` is enabled, because in that branch `screenshot_capturer.dart` draws *only* the `extractRects()` list — the separate `getPostHogWidgetWrapperElements()` list is drawn only in the `else` branch, when both flags are off. So with `maskAllTexts: true` there is no second path that recovers the dropped wrapper.

Every node below the root of this tree already matched a masking rule (`widget_elements_decipher.dart` only attaches nodes that a parser accepted), so no node may be discarded. The fix collects the whole subtree.

This affects **shipped iOS and Android** session replay. The masks are painted into the screenshot on the Dart side by `ImageMaskPainter`, so both platforms use the same code path and both leaked. The same function is what the in-progress web canvas mask provider calls, so it is fixed there too.

### Cost

Collecting the full subtree emits more rects, including exact duplicates — a `Text` element and the `RichText` below it resolve to the same `RenderParagraph`, so they produce an identical rect. Measured on the example app's masking screen: 39 rects before, 76 after, of which 38 are unique. These are `canvas.drawRect` calls of the same black rect on a Dart-side picture (nothing extra crosses the method channel), the output image is pixel-identical, and capture is throttled — so this is left un-deduplicated rather than adding a hash set to the per-frame path.

**Unit tests** (`posthog_flutter/test/element_data_test.dart`, new). Tree-shape tests for `extractRects` (deep nesting, flat trees, a node with several children, empty root), regression tests for `extractMaskWidgetRects`, plus widget-level tests that pump a real tree under `PostHogWidget` and assert against `PostHogMaskController.getCurrentWidgetsElements()` — the `PostHogMaskWidget`-with-multiple-children case, `RichText`, `SelectableText`, `Text`, `TextField`. Three of them fail on `main` and pass with the fix. Full suite: 234 passing, `flutter analyze` clean, `dart format --set-exit-if-changed` clean.

**iOS simulator (iPhone 17 Pro, iOS 26.4) and Android emulator (API 37)**, example app with `maskAllTexts: true`, before and after the fix, comparing the actual masked frames the SDK produces. On both platforms, before the fix: the `PostHogMaskWidget` around a row of `Text` / unmasked widget / `Text` masks only the two texts, and **the widget between them is fully visible in the recording**; the `PostHogMaskWidget` around the password field leaves its background and border uncovered. After the fix both wrappers are covered end to end on both platforms. (The simulator runs used a network image in that slot; the committed example uses a plain coloured box instead, so the screen has no network dependency — same masking behaviour, since neither matches a rule while `maskAllImages` is off.)

Checked for over-masking in the same runs: with `maskAllImages: false`, the standalone images in tests 6–8 stay unmasked, and rotated/scaled text masks still track their transforms. Texts, `RichText`, `SelectableText`, and text fields were already masked before the fix on mobile — that part was not a regression, and this change does not alter it.

Example app: added test 15 to the masking screen, a `PostHogMaskWidget` wrapping several masked children with an unmasked widget between them, which is the shape that leaked. Tests 2 and 3 now state the config their labels depend on, matching test 1 — both are masked only when `maskAllTexts` is on, so the old unconditional wording read like a bug under the example's committed defaults.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Written with Claude Code. The starting hypothesis was that `RichText` and `SelectableText` are never masked on mobile; that turned out to be wrong — both sit at the root of the matched-element tree in practice and were already masked. Dumping the real tree from the example app is what located the actual failure: the dropped `PostHogMaskWidget` wrapper. The example screen's test 15 and the widget-level test were written to reproduce that specific shape rather than the original hypothesis.

An earlier draft kept the "drop a node with more than one child" branch and only added recursion. That was discarded — with every node in the tree being a confirmed masking match, there is no rule that justifies dropping one, and keeping the branch left the wrapper leak in place. De-duplicating the resulting rects was considered and rejected on the measurement above.

Two adjacent pre-existing bugs were found and deliberately left alone:

- `PostHogMaskController.parsers` is `late final`, built once from the startup config, so changing `maskAllTexts` / `maskAllImages` at runtime never registers or unregisters the parsers.
- `element_object_parser.dart` masks `element.widget is Text` with no `maskAllTexts` check, so `maskAllImages: true` with `maskAllTexts: false` still masks all text.

Also noted, not fixed: `screenshot_capturer.dart` calls `getPostHogWidgetWrapperElements()` unconditionally and then `getCurrentWidgetsElements()`, parsing the whole render tree twice per captured frame, and discards the first result whenever `maskAllTexts` or `maskAllImages` is on.

### DIFF
--- a/posthog_flutter/lib/src/replay/element_parsers/element_data.dart
+++ b/posthog_flutter/lib/src/replay/element_parsers/element_data.dart
@@ -29,22 +29,15 @@ class ElementData {
     return elements;
   }
 
-  List<ElementData> extractRects({bool isRoot = true}) {
-    List<ElementData> rects = [];
+  /// Every node below the root is an element that already matched a masking
+  /// rule, so the whole subtree is collected — a match can sit at any depth
+  /// (a `SelectableText` puts its `RenderParagraph` several levels down).
+  List<ElementData> extractRects() {
+    final rects = <ElementData>[];
 
-    if (children != null) {
-      for (var child in children ?? []) {
-        if (child.children == null) {
-          rects.add(child);
-          continue;
-        } else if ((child.children?.length ?? 0) > 1) {
-          for (var grandChild in child.children ?? []) {
-            rects.add(grandChild);
-          }
-        } else {
-          rects.add(child);
-        }
-      }
+    for (final child in children ?? const <ElementData>[]) {
+      rects.add(child);
+      rects.addAll(child.extractRects());
     }
     return rects;
   }
--- a/posthog_flutter/lib/src/replay/element_parsers/element_data.dart
+++ b/posthog_flutter/lib/src/replay/element_parsers/element_data.dart
@@ -1,5 +1,3 @@
-// ignore_for_file: avoid_dynamic_calls
-
 import 'package:flutter/material.dart';
 import 'package:posthog_flutter/src/replay/mask/posthog_mask_widget.dart';
 
@@ -31,7 +29,8 @@ class ElementData {
 
   /// Every node below the root is an element that already matched a masking
   /// rule, so the whole subtree is collected — a match can sit at any depth
-  /// (a `SelectableText` puts its `RenderParagraph` several levels down).
+  /// (a `ListTile` title nests `AnimatedDefaultTextStyle` → `DefaultTextStyle`
+  /// → `Text` → `RichText`).
   List<ElementData> extractRects() {
     final rects = <ElementData>[];
 
--- a/.changeset/mask-nested-elements.md
+++ b/.changeset/mask-nested-elements.md
@@ -1,0 +1,5 @@
+---
+"posthog_flutter": patch
+---
+
+Session replay: masks every element that matched a masking rule, at any depth. The rect collector only walked two levels of the matched-element tree and dropped any node that had more than one matched child, so with `maskAllTexts`/`maskAllImages` enabled a `PostHogMaskWidget` wrapping several masked children lost its own mask rect and anything it wrapped but did not match on its own (an image, an avatar, decoration) stayed visible in the recording.
--- a/.changeset/mask-nested-elements.md
+++ b/.changeset/mask-nested-elements.md
@@ -2,4 +2,4 @@
 "posthog_flutter": patch
 ---
 
-Session replay: masks every element that matched a masking rule, at any depth. The rect collector only walked two levels of the matched-element tree and dropped any node that had more than one matched child, so with `maskAllTexts`/`maskAllImages` enabled a `PostHogMaskWidget` wrapping several masked children lost its own mask rect and anything it wrapped but did not match on its own (an image, an avatar, decoration) stayed visible in the recording.
+Session replay: masks every element that matched a masking rule, at any depth. The rect collector only walked two levels of the matched-element tree and dropped any node that had more than one matched child, so with `maskAllTexts`/`maskAllImages` enabled a `PostHogMaskWidget` wrapping several masked children lost its own mask rect, and whatever it wrapped that matched no rule on its own (a decorated container, a chart, a custom-painted avatar) stayed visible in the recording. More of the screen is masked as a result: a `PostHogMaskWidget` now covers its whole subtree, as documented, instead of only the children that matched.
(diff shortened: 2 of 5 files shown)


posthog_flutter/lib/src/replay/element_parsers/element_data.dart
```diff
@@ -29,22 +29,15 @@ class ElementData {
...
+  /// Every node below the root is an element that already matched a masking
+  /// rule, so the whole subtree is collected — a match can sit at any depth
+  /// (a `SelectableText` puts its `RenderParagraph` several levels down).
+  List<ElementData> extractRects() {
+    final rects = <ElementData>[];
+
+    for (final child in children ?? const <ElementData>[]) {
+      rects.add(child);
+      rects.addAll(child.extractRects());
     }
```
dustinbyrne, line 40: any idea why we previously limited to a depth of two? any considerations regarding performance or recursion depth here?

dustinbyrne (approved):

posthog_flutter/lib/src/replay/element_parsers/element_data.dart
turnipdabeets: Good question — I went digging, and the honest answer is that it looks unfinished rather than deliberate.

`extractRects` landed in #123 with the shape of a recursive function: `extractRects([bool isRoot = true])` opened with `if (!isRoot) rects.add(rect);`. But the body never recursed — it hand-walked two levels. #147 then dropped that `!isRoot` self-add when the return type changed from `List<Rect>` to `List<ElementData>`, which left `isRoot` an unused parameter ever since (this PR removes it). No commit records a performance rationale for the depth cap; the one perf change in this file, #269, optimized the sibling `extractMaskWidgetRects` and deliberately left this one alone.

**Recursion depth:** bounded by *matched* nodes, not widget-tree depth — only elements that matched a masking rule become nodes. The deepest real chain I could produce is 4 (the `ListTile` title chain in the comment). `WidgetElementsDecipher._parseAllElements` already recurses the entire, far deeper `Element` tree inside the same `try`, so this adds no overflow surface that wasn't already there. Measured: depth 20,000 completes in ~1s, 100,000 throws `StackOverflowError`, which `getCurrentWidgetsElements()`'s `catch (e)` does catch on the Dart VM.

**Performance:** the walk itself is unchanged in cost — `getCurrentWidgetsElements()` is dominated by `parseRenderTree`, which is identical. What grows is the rect count: 39→76 on the example masking screen, 20→40 for 200 plain `Text` rows, 18→72 for 200 `ListTile`s, 600→1200 for 600 `Text`s. That's ~0.2ms extra in `ImageMaskPainter.drawMaskedImage` at 1200 rects, on a path throttled to roughly 1fps.

Nearly all of the growth is exact duplicates: composition-only ancestors (`Text`, `RichText`, `DefaultTextStyle`, `Expanded`) resolve `element.renderObject` to the *same* `RenderParagraph`, so they carry an identical rect — 76 rects on the example screen, 38 unique. They can't change the output: opaque black, `PaintingStyle.fill`, default `srcOver`, no `saveLayer`, so over-draw is idempotent.

I did consider stopping descent at the first match (roughly what android/ios do) to avoid the duplicates, and rejected it — a matched descendant can paint *outside* its matched ancestor's `paintBounds`. With `Transform.translate`, `OverflowBox`, and `Stack(clipBehavior: Clip.none)` + `Positioned`, the descendant text ends up completely unmasked in all three. Android gets away with it because `getGlobalVisibleRect()` clips, which is a different mechanism. Deduping by exact rect is safe but buys ~0.2ms, so I left it out.

PR #501, opened 2026-07-27
feat(replay): PostHogMaskWidget enables web canvas masking on its own

turnipdabeets:
> **Stacked on #499** (`feat/web-canvas-masking`) — this PR targets that branch, not `main`. Review it after #499, or squash the two together before merging; they are one feature.

#499 gives Flutter web canvas masking, but it only registers the mask-region provider with posthog-js when the app already declared `session_recording.canvasCapture.maskRegionsFn` in its `posthog.init` call.

That leaves a footgun: a developer wraps sensitive UI in `PostHogMaskWidget`, does not also edit `web/index.html`, and gets **zero masking on web, silently**. `PostHogMaskWidget` is the most explicit possible statement of "never record this", and it needs no setup at all on iOS and Android — so web quietly diverging is the worst possible failure mode.

The gate itself is not wrong. Registering does two things beyond attaching a function: it adds `blockSelector: 'flt-semantics-host'` (so Flutter's accessibility DOM is not recorded in the clear), and it restarts an in-flight recording, because posthog-js only reads those options when rrweb's `record()` starts. Imposing that on every Flutter web app — including ones that never asked for masking — is not acceptable.

So this PR adds a second way to opt in rather than removing the gate: **a mounted `PostHogMaskWidget` is itself the opt-in.**

## :hammer: What changed

- `PostHogMaskWidgetState.initState` calls `notifyMaskWidgetMounted()`, resolved through a conditional import (`canvas_mask_registration_io.dart` / `..._web.dart`). Off web it is an empty function body, so no web code reaches mobile builds and nothing changes on iOS/Android/macOS.
- On web the hook defers to the end of the frame and calls `WebCanvasMaskProvider.notifyMaskWidgetMounted()`. Deferring matters: registration restarts the recording, and rrweb can take a canvas snapshot from inside that call — which would walk a widget tree that is still being built.
- `WebCanvasMaskProvider` keeps the same "did the app declare it?" gate, but now also applies when a mask widget has mounted. The retry/apply path became a small state machine (`applied` / `notOptedIn` / `posthogNotReady`) so that "posthog-js is up but the app did not opt in" parks instead of terminating — a `PostHogMaskWidget` mounting later re-enters it. Registration is idempotent: at most one `set_config` and at most one recording restart, however many mask widgets mount, and whether posthog-js loads before or after Flutter.

Apps with no `PostHogMaskWidget` and no declaration are still completely untouched: no `set_config`, no `blockSelector`, no restart.

Also drops a stale caveat from the example app: test 15's section title claimed `PostHogMaskWidget` with multiple children "needs maskAllTexts or maskAllImages". It does not, on either platform — `getMaskElements` calls `extractMaskWidgetRects()` unconditionally on web, and `screenshot_capturer` calls `getPostHogWidgetWrapperElements()` outside the flags branch on mobile (post-#500). A browser check against the built app on `feat/web-canvas-masking` with `maskAllTexts: false, maskAllImages: false` confirmed it: the provider returned a full-width rect covering the whole Row, amber box included.

## :warning: Cost and residual caveats

These are documented in the CHANGELOG and in the `PostHogMaskWidget` dartdoc:

- **One mount enables the whole masking configuration, not just the wrapped subtree**: `maskAllTexts`/`maskAllImages` default to true, so a single `PostHogMaskWidget` turns on full text and image canvas masking — the same semantics as mounting one on iOS/Android.
- **The recording restarts once, mid-session,** the first time a mask widget mounts. The replay will show a split at that point. Unavoidable without a posthog-js change: `blockSelector` is a start-time option.
- **Frames captured before the first mount are recorded unmasked.** Declaring `maskRegionsFn: () => null` in `posthog.init` is still the only thing that covers the window between page load and Flutter booting (those frames are skipped instead), and it also avoids the restart. The HTML setup is now an optimization, not a requirement.
- **`PostHogWidget` is still required.** An app that mounts a `PostHogMaskWidget` but is not wrapped in `PostHogWidget` now fails closed — canvas frames are skipped rather than recorded unmasked, with a console warning explaining the fix. That is the correct behavior for something that explicitly asked for masking, but it is a behavior change for that (misconfigured) shape of app. The outside-tree check runs once, at the mask widget's first mount: a mask widget that mounts before any `PostHogWidget` exists is treated as this no-`PostHogWidget` shape and does opt in (documented in the changeset, pinned by a test).

### By design: registration does not wait on `recordCanvas`

Registration opts in without first checking that canvas capture is actually enabled. That is deliberate, for two reasons:

- **Gating on it would fail open.** `recordCanvas` can arrive from remote config after `posthog.init`, so a `ph.config` check at registration time can read "canvas off" for a session that goes on to record the canvas — masking would silently never attach.
- **Registration earns its keep even with canvas capture off.** Applying the config also installs `blockSelector: 'flt-semantics-host'`, and that exclusion is independent of the canvas: with accessibility active, Flutter mirrors widget text into `flt-semantics` DOM nodes, which rrweb records as plaintext during ordinary DOM recording. So a canvas-off app that mounts a `PostHogMaskWidget` still gains a real thing — the a11y text side-channel is closed.

The restart is gated on `ph.sessionRecordingStarted()` regardless, so an app that is not recording pays nothing at all.

Three new tests in `test/web_canvas_mask_provider_test.dart` (already listed in the "Test (web)" CI job, so no workflow change needed):

- a mounted `PostHogMaskWidget` opts the app in when `posthog.init` declared nothing — provider attached, `blockSelector` set, recording restarted once;
- exactly one `set_config` and one restart across two pumps with three and then five mask widgets;
- a mask widget that mounts while posthog-js is still absent opts in as soon as posthog-js appears.

A follow-up commit (review fixes, merged with the updated #499 base) adds two more: cross-path idempotency (app opted in via the init declaration, then a mask widget mounts — still exactly one `set_config` and one restart), and a `set_config` throw during the mount-triggered apply being retried and applied on a later tick instead of permanently consuming the opt-in.

The pre-existing `leaves posthog-js untouched when the app declares no mask provider` test is unchanged and still passing, which is the assertion that apps opting into nothing stay untouched.

Ran locally, all green:

- `flutter test` — 247 passing
- `flutter test --platform chrome test/posthog_flutter_web_handler_test.dart test/web_canvas_mask_provider_test.dart` — 46 passing
- `flutter analyze` — no issues
- `dart format --output=none --set-exit-if-changed ./` — exit 0

The registration change itself was not manually verified in a browser against a real posthog-js build; the posthog-js side of the contract is exercised by the stub, as in #499. The test 15 measurement quoted above was taken by hand on the base branch, not by the agent.

## :pencil: Checklist

### If releasing new changes

- [x] Release note shipped as a one-line changeset (`.changeset/mask-widget-web-canvas.md`) per the changelog style; the behavioral detail (whole-config opt-in, one-time restart, pre-mount window, `PostHogWidget` requirement, mount-time outside-tree check) lives in this PR body and the `PostHogMaskWidget` dartdoc.

## 🤖 Agent context

Written with Claude Code (Opus 5). The design question was how to let shared widget code trigger web-only registration without leaking `dart:js_interop` into mobile builds; the repo's existing `_io.dart` / `_web.dart` conditional-import convention (see `origin.dart`, `chunk_ids.dart`) handles it, so the mount hook is a one-line no-op function on every non-web platform.

Two alternatives were rejected: removing the gate outright (would restart recording and drop the semantics tree for every Flutter web app, including ones that never asked for masking), and gating registration on `recordCanvas` being visibly enabled in `ph.config` (fails open when canvas capture arrives via remote config, and would forfeit the semantics exclusion for canvas-off apps — see "By design" above).

Deferring the hook to a post-frame callback was a deliberate change after the first working version called it synchronously from `initState` — registering synchronously means posthog-js may re-enter the mask-region provider mid-build, which walks the render tree.


posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -185,22 +243,25 @@ class WebCanvasMaskProvider {
     final config = JSObject();
     config.setProperty('session_recording'.toJS, sessionRecording);
     ph.set_config(config);
+    _applied = true;
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — fixed in 04c4bcd: `_applied` is now set only after the stop/start block, and `register()`'s catch schedules the retry chain instead of ending it, so a restart that throws after `set_config` landed gets retried until the selector actually takes effect (regression test: stub whose `stopSessionRecording` throws once → full apply retried, restart completes; verified the test fails without the reorder).

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -89,9 +122,28 @@ class WebCanvasMaskProvider {
...
-  void _registerUnsafe() {
-    final ph = posthog;
-    if (ph != null && _tryApplyConfig(ph)) {
+  void _onMaskWidgetMounted() {
+    if (_maskWidgetMounted) {
+      return;
+    }
+    _maskWidgetMounted = true;
+    // a retry chain still in flight picks the flag up on its next tick
+    if (_applied || _polling) {
```
marandaneto, line 175: When this apply throws, the retry is scheduled without setting `_polling = true`. If a `PostHogMaskWidget` mounts before that timer fires, it passes the `_polling` guard and starts a second retry chain. Since only the latest timer is retained, a later setup cancels one chain while the sibling can survive and reapply the old provider/config. I reproduced this sequence locally. Could we mark polling active here and enforce at most one outstanding retry timer?

posthog_flutter/lib/src/replay/mask/posthog_mask_widget.dart
```diff
@@ -27,6 +38,7 @@ class PostHogMaskWidgetState extends State<PostHogMaskWidget> {
   @override
   void initState() {
     super.initState();
+    notifyMaskWidgetMounted();
```
marandaneto: This globally opts in without verifying that the mounted mask widget belongs to the `PostHogWidget` subtree used by `PostHogMaskController`. If a different `PostHogWidget` exists as a sibling, the walk succeeds on that other subtree and returns an empty/incomplete region list instead of failing closed, leaving this explicitly masked widget exposed. I reproduced the sibling-tree case locally. Could we validate the triggering widget's ancestry or otherwise fail closed when it is outside the tracked tree?

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
```diff
@@ -190,20 +248,25 @@ class WebCanvasMaskProvider {
...
+    // since the semantics exclusion applies to plain DOM recording too (and
+    // recordCanvas can arrive from remote config after init anyway)
     if (ph.sessionRecordingStarted()) {
       ph.stopSessionRecording();
       ph.startSessionRecording();
     }
-    return true;
+    // only after the restart: a throw above must leave the apply retryable,
+    // or the selector never takes effect for this recording
+    _applied = true;
```
marandaneto, line 315: If `stopSessionRecording()` succeeds but `startSessionRecording()` throws, the retry sees recording as stopped, skips this block, and marks the provider applied, leaving recording permanently stopped. The current test only throws from `stopSessionRecording()` and its stub does not model recording-state transitions. Could we retain a pending-restart state and retry the start independently?

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
turnipdabeets: Fixed in d1a01b4 — `_scheduleRetry` now owns the invariant: it cancels any existing timer and sets `_polling = true` itself, so every retry path is marked and at most one timer exists. Test: throwing `set_config` + a mount before the retry + a second `register()` → exactly one successful `set_config`/restart, no orphaned chain across a 5s pump.

posthog_flutter/lib/src/replay/mask/posthog_mask_widget.dart
turnipdabeets: Fixed in e1900cd — `notifyMaskWidgetMounted` now carries the mounting `BuildContext`, and the web hook refuses the opt-in when a tracked tree exists and the widget's render object is provably outside it. No tracked tree at all still opts in, preserving the documented fail-closed-with-warning path for that shape. Tests cover outside/inside/no-tree.

posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart
turnipdabeets: Fixed via the same pending-restart mechanism from the #499 commit 7b8480d, merged in 02740e4 so `_applied` stays false until the restart fully succeeds; the stateful-stub test runs against this branch's state machine too.

turnipdabeets:
valid — the "retry loop never backs off when _apply() keeps throwing" finding was correct (confirmed the exception path reused the incoming delay while the posthog-not-ready path ramped to 4s, and the same structure existed on the base branch #499). Fixed in 017c8a1: the backoff now computes up front so both paths follow the same 250ms→4s curve, with a regression test discriminating the doubling chain from fixed-rate retries by attempt count. Nice catch on the reachability analysis too — it correctly bounded the impact to log spam rather than data loss.

.changeset/mask-widget-web-canvas.md
```diff
@@ -0,0 +1,35 @@
...
+- Your app must be wrapped in `PostHogWidget`. If it is not, canvas frames are
+  skipped instead of recorded unmasked, and a console warning explains the fix.
+- A `PostHogMaskWidget` mounted outside the `PostHogWidget` tree does not opt the
+  app in: masking could never cover it, so opting in would expose it instead of
+  failing closed. A debug log explains the misplacement. This check runs once, at
+  the mask widget's first mount: a mask widget that mounts before any
+  `PostHogWidget` exists is treated as the no-`PostHogWidget` shape above and
+  does opt in.
+
+Apps that declare neither are untouched, exactly as before.
```
marandaneto: changelog should be small and concise, add this info to docs if needed and link to it if needed

posthog_flutter/lib/src/replay/mask/canvas_mask_registration_web.dart
```diff
@@ -0,0 +1,66 @@
...
+/// app missing PostHogWidget.
+///
+/// The check runs once, in the mount's post-frame callback: a null tracked
+/// context at that moment is treated as the no-PostHogWidget shape and
+/// allowed, even if a PostHogWidget later mounts without containing this
+/// widget.
+bool _isInTrackedTree(BuildContext context) {
+  final trackedContext =
+      PostHogMaskController.instance.containerKey.currentContext;
+  if (trackedContext == null) {
```
marandaneto, line 57: blocking: This one-time null-tree allowance can fail open after a later rebuild. If this `PostHogMaskWidget` mounts now, the opt-in is latched; if a sibling `PostHogWidget` mounts later, `getMaskElements` succeeds against only that sibling tree and returns empty/incomplete regions, so the explicitly masked subtree can be recorded unmasked. The existing before/after test preserves this ordering but only checks registration counts, not the emitted regions. Could we keep failing closed unless this context becomes a descendant of the tracked tree, or revalidate it when regions are computed?

marandaneto (approved):
left a last comment, approving to unblock

.changeset/mask-widget-web-canvas.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog_flutter": minor
```
marandaneto: maybe we dont need this changeset at all since its gonna be merged to another branch?

turnipdabeets: Yea my plan was to put them together on the merge.

posthog_flutter/lib/src/replay/mask/canvas_mask_registration_web.dart
turnipdabeets: Fixed in 7eec532 — took the revalidate-at-region-computation direction since it covers the whole class, not just this ordering. Mask widgets register on mount / unregister on dispose, and every `maskRegionsFn` call verifies all mounted mask widgets are inside the tracked tree (outside the per-frame cache, so a stale value can't survive a tree change). Any outsider → the frame is skipped with a one-time debug log; removing or reparenting it recovers. Your scenario's test now asserts the emitted regions are null, plus new tests for the late-`PostHogWidget`-contains-it shape and recovery after removal.

posthog_flutter/lib/src/replay/mask/canvas_mask_registration_web.dart
```diff
@@ -0,0 +1,66 @@
...
+/// The masking walk only sees PostHogWidget's subtree, so a mask widget
+/// outside it would opt masking in while its own rects are never produced —
+/// the walk would succeed and ship rects that do not cover the widget. With
+/// no tracked tree at all the opt-in stays allowed: every walk then fails and
+/// frames are skipped (fail closed), which is the documented behavior for an
+/// app missing PostHogWidget.
+///
+/// The check runs once, in the mount's post-frame callback: a null tracked
+/// context at that moment is treated as the no-PostHogWidget shape and
+/// allowed, even if a PostHogWidget later mounts without containing this
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — fixed in e89ff8e. The gate (and the per-frame revalidation, which had the same mismatch) now resolves the tracked-tree root exactly the way `RootElementProvider` does: active `ModalRoute` on the tracked context → the root navigator's render object, else the tracked context's own. Both membership checks share the new `WebCanvasMaskProvider.trackedTreeRoot()`, so a `PostHogMaskWidget` in a root-navigator dialog is tracked (opts in, produces regions) instead of being rejected — pinned by a new dialog-shape test. The Navigator/ModalRoute lookups fall back to the tracked context's render object if they throw.

posthog_flutter/lib/src/replay/mask/canvas_mask_registration_web.dart
```diff
@@ -0,0 +1,66 @@
...
+      printIfDebug(
+        'PostHog: this PostHogMaskWidget is outside the PostHogWidget tree '
+        'PostHog tracks, so masking could never cover it — it does not '
+        'enable web canvas masking.',
+      );
+      return;
+    }
+    WebCanvasMaskProvider.notifyMaskWidgetMounted();
+  });
+}
```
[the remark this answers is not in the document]
turnipdabeets: Fixed in e89ff8e — the callback body is now try/catch-wrapped with a `printIfDebug`, matching the sibling hops, so a throw stays inside the SDK's debug logging instead of reaching `FlutterError.onError` (and the host's error tracking as `handled: false`). Fair hit on our own triage: we had deferred this exact guard as "no demonstrated throw path" — the error-dashboard consequence is the right reason to add it regardless.

posthog_flutter/lib/src/replay/mask/canvas_mask_registration_web.dart
turnipdabeets: CI is green — resolving. Final shape after the ReviewHog pass: the revalidation (and the mount gate) now resolve the tracked-tree root the same route-dependent way the masking walk does (`trackedTreeRoot()`, e89ff8e), so a mask widget in a root-navigator dialog is tracked rather than blanking frames, while the sibling-tree fail-open you flagged stays closed — pinned by the dialog-shape and outside-tree tests.

PR #502, opened 2026-07-27
perf(replay): skip screenshot capture on Flutter web

turnipdabeets:

On Flutter **web**, `PostHogWidget` runs the **mobile** session replay screenshot pipeline and throws away every frame it produces.

**Evidence the output is discarded.** Web replay is recorded by posthog-js, not by this pipeline, and the SDK already knows it in two places:

- `NativeCommunicator.isSessionReplayActive()` starts with `if (kIsWeb) { return false; }` — the gate `captureScreenshot()` awaits before rasterizing is hardcoded false on web.
- Even if it weren't, `handleWebMethodCall` no-ops the delivery methods: `case 'sendMetaEvent': // not supported on Web` and `case 'sendFullSnapshot': // not supported on Web`.

**Where the cost is.** It isn't the discarded PNG — the pipeline never gets that far. It's the two synchronous render-tree walks that happen *before* the gate. `captureScreenshot()` calls `PostHogMaskController.instance.getPostHogWidgetWrapperElements()` unconditionally, and `getCurrentWidgetsElements()` when `maskAllTexts || maskAllImages` (both default `true`), and only then enters the `Future(() async { ... await isSessionReplayActive() ... })` that bails. Each walk is a full `parseRenderTree` over the subtree under `PostHogWidget` plus a rect extraction.

`ChangeDetector` is a 1s `Timer.periodic` that also `addPostFrameCallback`s each tick. So a web app with `config.sessionReplay = true` and `PostHogWidget` mounted does two whole render-tree walks per second, forever, and drops the result every time.

This is pre-existing on `main`, not a regression from any in-flight branch.

## :hammer: What changed

One place, two `kIsWeb` guards, both in **`PostHogWidget`**: `initState` no longer starts the detector on web, and `_startRecording()` (reached from the `sessionRecordingActive` listener, e.g. via `Posthog().startSessionRecording()`) returns early. That removes the timer and the per-frame callbacks entirely, not just the walks — nothing downstream ever runs, so the walks never happen either.

An earlier revision also bailed inside `ScreenshotCapturer.captureScreenshot()`. That is gone. It was unreachable in production — `ScreenshotCapturer` is not exported from `lib/posthog_flutter.dart` and is only constructed in `_initComponents`, which the widget guard already blocks on web — and worse, it was masking the tests: because it returned before scheduling any work, deleting the `PostHogWidget` guard left the whole web suite green. Removing it is what makes the tests fail when the guard that matters is touched.

`build()` is untouched. `PostHogWidget` still mounts the `RepaintBoundary` keyed by `PostHogMaskController.instance.containerKey`, so anything that resolves a context through that key — including the in-flight web canvas-masking work — behaves exactly as before. There is a test pinning that.

No new public API. `posthog_flutter.dart` exports `src/posthog_widget.dart` wholesale, so any member added to `PostHogWidgetState` lands in `api/posthog_flutter.api.json` (the snapshot script does not filter `@visibleForTesting`). The tests assert observables instead, and the snapshot is unchanged from `main`.

Secondary effect: `kIsWeb` is a dart2js compile-time constant, so both guarded branches are statically dead on web. `ScreenshotCapturer` and `ChangeDetector` become unreachable from `PostHogWidget` there and should tree-shake out of web bundles, along with the capture and mask-parsing code they pull in. Not measured — reported as a mechanism, not a byte count.

## :mag: What I verified is *not* depended on

The risk here is that skipping `_initComponents` on web leaves something else reading a now-null field, or that some other path needs `_startRecording` to have run. I traced all of it:

- **Field null-safety.** `_screenshotCapturer`, `_nativeCommunicator`, `_changeDetector` are accessed exclusively through `?.` at every one of their ~15 use sites in `posthog_widget.dart`, including `_setSuppressFlutterCapture`, `_onNativeOcclusionChanged`, `_generateSnapshot`, `_sendSnapshot`, `_stopRecording` and `dispose`. Leaving them null on web is already a supported state (it is the state before the first `_initComponents`).
- **Occlusion path.** `_onNativeOcclusionChanged` is the only other consumer of those fields. It fires off `PostHogInternalEvents.nativeOcclusionEvent`, which is pushed from exactly one place: the `onNativeOcclusionChanged` method-call case in `posthog_flutter_io.dart`. On web the platform instance is `PosthogFlutterWeb`, so that handler is never installed and the event never fires. The handler is null-safe regardless.
- **No other entry into the capture path.** `ScreenshotCapturer` and `ChangeDetector` are constructed only in `PostHogWidget._initComponents`; `captureScreenshot()` has exactly one production call site (`_generateSnapshot`). There is no public `Posthog().captureScreenshot()` API. `PosthogObserver` does not touch replay capture at all.
- **Mask controller.** The only production callers of `getCurrentWidgetsElements()` / `getPostHogWidgetWrapperElements()` are inside `captureScreenshot()`. `containerKey` is read by the element parsers, which run only underneath those calls — plus `build()`, which is unchanged.
- **Nothing latches state from a first capture.** The only cross-object state written by the pipeline is `_changeDetector.hasCapturedPlatformViews`, set from `_generateSnapshot` and read only by `ChangeDetector._scheduleFrameCallback`. Both are gone together on web.

Nothing depends on this running on web.

- New `posthog_flutter/test/posthog_widget_web_test.dart` (`@TestOn('browser')`, run under `--platform chrome`): mounting `PostHogWidget` with `sessionReplay = true` leaves no capture work pending; flipping `sessionRecordingActive` to true keeps it that way; and the mask controller's `containerKey` still resolves a live context.
- **These three tests were run against unpatched `main`, where all three fail.** That is the confirmation that the waste is real and pre-existing, not inferred.
- How the first two assert, since nothing on web observes the pipeline: `captureScreenshot()` posts a zero-duration timer via `Future(...)` that no cleanup cancels (`dispose` only sets the capturer's cancelled flag), and `flutter_test` fails any test still holding a pending timer. Not the 1s `Timer.periodic` — `flutter_test` unmounts the tree before checking invariants, which runs `dispose()` and cancels that one. It only trips if the clock is not advanced, so these tests must not pump a duration; with `pump(const Duration(seconds: 2))` they pass on `main` and catch nothing. Both facts are recorded in the file's header comment.
- **Mutation-tested per guard**, because "fails on `main`, passes here" only proves the guards are load-bearing *together*. Each guard was removed or inverted individually against the final tree: `initState`'s guard removed → all 3 web tests fail; `_startRecording`'s guard inverted → the new VM restart test fails. Both pass with the tree intact.
- That pass is what caught the `_startRecording` guard having **no coverage on any platform**: inverting it to `if (!kIsWeb) return;` — so mobile capture can never resume after `stopSessionRecording()` / `startSessionRecording()`, freezing replay on the last frame with no error — passed all 236 tests. `resumes capture after recording is toggled off and on` now fails on that mutant.
- Two VM tests in `native_occlusion_episode_test.dart`: capture *does* reach `captureScreenshot()` off web (via the `isSessionReplayActive` channel call it makes), and it resumes after a recording restart. Neither can pass if the guard leaks off web.
- Added the new file to the "Test (web)" step in `ci.yml`, which lists web test files explicitly.
- `flutter test`: 237 passing, up from 235 on `main` (two added VM tests); no regressions. `flutter test --platform chrome` on both web files, as CI runs them: 11 passing. `dart analyze`: clean. `dart format --set-exit-if-changed`: clean. `make checkApiDart`: snapshot unchanged.
- Not run: no manual browser session against a real web app; the cost argument is from reading the call path plus the failing-then-passing tests above, not from a profiler trace.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Claude Code (Opus 5) did the tracing and implementation. The starting suggestion was to guard only `PostHogWidget`, which is where this ended up. I deliberately did not restructure `captureScreenshot()` to move the `isSessionReplayActive()` await ahead of the walks — that would change the ordering semantics of the whole capture for mobile too, which is a bigger change than this PR should carry.

The verification order was chosen to avoid asserting the problem existed: the web tests were written and run against unmodified `main` first, so their failure is the evidence, and the fix is what turns them green.

Three things were wrong along the way, all caught by re-running rather than by reasoning, and worth recording because each produced a green suite that proved nothing:

1. The tests first read a new `debugCaptureRunning` getter on `PostHogWidgetState`. That worked, but grew the public API snapshot — `src/posthog_widget.dart` is exported wholesale and the check script does not filter `@visibleForTesting`. Replaced with the pending-timer assertion.
2. The replacement pumped a 2s duration to "let the timer fire", which *flushes* the timer. Two of the three tests then passed against unpatched `main`, detecting nothing.
3. Even once green-on-branch and red-on-`main`, the suite only pinned the guards **as a pair**: the `captureScreenshot()` bail returned before scheduling the timer the tests watch, so deleting the `PostHogWidget` guard — the entire point of the PR — kept all three green. Found by mutating each guard separately. Dropping that bail fixed it, and the same mutation pass surfaced the untested `_startRecording` guard.

The lesson worth carrying: "fails before the fix, passes after" is necessary but not sufficient when a change has more than one part. Only per-part mutation showed which assertions were real.

https://claude.ai/code/session_01HPiKky15SKqNg9PT2wHeaw

### DIFF
--- a/.changeset/web-skip-screenshot-capture.md
+++ b/.changeset/web-skip-screenshot-capture.md
@@ -1,0 +1,5 @@
+---
+"posthog_flutter": patch
+---
+
+Flutter web: `PostHogWidget` no longer runs the mobile session replay screenshot pipeline. On web, session replay is recorded by posthog-js, and every snapshot this pipeline produced was discarded — but it still walked the whole render tree twice per second (once for `PostHogMaskWidget` wrappers, once more when `maskAllTexts` or `maskAllImages` is on) for as long as the app was open. Web apps that enable `sessionReplay` and mount `PostHogWidget` no longer pay for that. Replay on web is unaffected, and `PostHogWidget` still mounts as before, so it remains safe to keep in a shared widget tree across web and mobile.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,7 +167,7 @@ jobs:
       - name: Test (web)
         if: needs.detect-markdown-only.outputs.markdown_only != 'true'
         working-directory: ./posthog_flutter
-        run: flutter test --platform chrome test/posthog_flutter_web_handler_test.dart
+        run: flutter test --platform chrome test/posthog_flutter_web_handler_test.dart test/posthog_widget_web_test.dart
 
   publish-dry-run:
     needs: detect-markdown-only
--- a/api/posthog_flutter.api.json
+++ b/api/posthog_flutter.api.json
@@ -4698,17 +4698,6 @@
                         "name": "debugFlutterCaptureSuppressed",
                         "relativePath": "lib/src/posthog_widget.dart",
                         "typeName": "bool"
-                    },
-                    {
-                        "entryPoints": [],
-                        "isDeprecated": false,
-                        "isExperimental": false,
-                        "isReadable": true,
-                        "isStatic": false,
-                        "isWriteable": false,
-                        "name": "debugCaptureRunning",
-                        "relativePath": "lib/src/posthog_widget.dart",
-                        "typeName": "bool"
                     }
                 ],
                 "isDeprecated": false,
--- a/api/posthog_flutter.api.json
+++ b/api/posthog_flutter.api.json
@@ -4698,6 +4698,17 @@
                         "name": "debugFlutterCaptureSuppressed",
                         "relativePath": "lib/src/posthog_widget.dart",
                         "typeName": "bool"
+                    },
+                    {
+                        "entryPoints": [],
+                        "isDeprecated": false,
+                        "isExperimental": false,
+                        "isReadable": true,
+                        "isStatic": false,
+                        "isWriteable": false,
+                        "name": "debugCaptureRunning",
+                        "relativePath": "lib/src/posthog_widget.dart",
+                        "typeName": "bool"
                     }
                 ],
                 "isDeprecated": false,
--- a/posthog_flutter/lib/src/posthog_widget.dart
+++ b/posthog_flutter/lib/src/posthog_widget.dart
@@ -43,9 +43,6 @@ class PostHogWidgetState extends State<PostHogWidget> {
   @visibleForTesting
   bool get debugFlutterCaptureSuppressed => _suppressFlutterCapture;
 
-  @visibleForTesting
-  bool get debugCaptureRunning => _changeDetector?.isRunning ?? false;
-
   void _setSuppressFlutterCapture(bool value) {
     _suppressFlutterCapture = value;
     _changeDetector?.suppressForcedFrames = value;
@@ -60,8 +57,8 @@ class PostHogWidgetState extends State<PostHogWidget> {
       return;
     }
 
-    // On web, session replay is recorded by posthog-js; the screenshot pipeline
-    // has no consumer there, so starting it would only burn frames.
+    // On web, session replay is recorded by posthog-js; this pipeline has no
+    // consumer there, so every snapshot it produces is discarded.
     if (!kIsWeb && config.sessionReplay) {
       _initComponents(config);
       _changeDetector?.start();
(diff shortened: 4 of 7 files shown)


turnipdabeets:
@greptileai review

ioannisj (approved):

PR #507, opened 2026-07-30
fix(replay): honor changed masking flags on iOS/Android across setup() calls

turnipdabeets:

Follow-up to the review note on #499 ([thread](https://github.com/PostHog/posthog-flutter/pull/499#discussion_r3683789350)): `refreshParsers()` was only wired into the web setup path, so on iOS/Android a second `setup()` with changed `maskAllTexts`/`maskAllImages` kept masking replay screenshots from the stale parser map built at first singleton access — e.g. `maskAllImages: false → true` would leave images unmasked (a narrow PII-leak direction). Pre-existing on `main`, not a regression from #499.

Two changes close it:

1. **Shared `setup()` refreshes the parser map** — `PostHogMaskController.instance.refreshParsers(config.sessionReplayConfig)` now runs on every platform. The web-side call in `WebCanvasMaskProvider.register()` stays: it is idempotent, and keeps the provider correct on its own if its entry point ever changes.
2. **`ScreenshotCapturer` resolves the live config at capture time** (`Posthog().config ?? _config`). A device test of change 1 alone showed masking still stuck: the capturer is built once per widget lifecycle and gated the mask walk on the `maskAllTexts`/`maskAllImages` flags of the config object captured at construction, so a second `setup()`'s new config never reached it — the parser map refreshed, but the walk was skipped before consulting it.

#499 merged while this was being opened, so this targets `main` directly (rebased to a single commit on top of the #499 squash merge).

- New tests: a second `setup()` with different masking flags rebuilds the parser map, and the capturer resolves the config of the latest `setup()` (both fail without their fix — verified by reverting each and re-running).
- Full suite: 249 tests pass, `flutter analyze` and `dart format` clean.
- Device E2E on an Android emulator: recorded a session with `maskAll*: false`, triggered an in-app second `setup()` with `maskAll*: true`, and confirmed in the PostHog replay that frames flip from readable to fully masked at the re-setup point. The same run with only change 1 left phase 2 unmasked, which is how change 2 was found.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

- Claude Code session: scoped from the PostHog review-bot finding on #499 (`consider`/`bug`), which validated the gap and its trigger conditions; the fix follows the bot's suggested wire-up point.
- Chose to keep the web-only `register()` call rather than remove it: removal would touch reviewed #499 code for no behavior change, and the duplicate refresh is idempotent.
- Refresh is unconditional (not gated on `config.sessionReplay`) so toggling replay flags across `setup()` calls always lands; the singleton instantiation this forces is a few small allocations.

### DIFF
--- a/posthog_flutter/test/posthog_test.dart
+++ b/posthog_flutter/test/posthog_test.dart
@@ -6,6 +6,7 @@ import 'package:posthog_flutter/posthog_flutter.dart';
 import 'package:posthog_flutter/src/posthog_flutter_platform_interface.dart';
 import 'package:posthog_flutter/src/posthog_internal_events.dart';
 import 'package:posthog_flutter/src/replay/mask/posthog_mask_controller.dart';
+import 'package:posthog_flutter/src/replay/screenshot/screenshot_capturer.dart';
 
 import 'posthog_flutter_platform_interface_fake.dart';
 
@@ -56,6 +57,21 @@ void main() {
         await Posthog().setup(imagesUnmasked);
         expect(controller.parsers.keys, isNot(contains('RenderImage')));
         expect(controller.parsers.keys, contains('RenderParagraph'));
+      },
+    );
+
+    test(
+      'screenshot capturer resolves the live config after a second setup',
+      () async {
+        final first = PostHogConfig('test_project_token')
+          ..sessionReplayConfig.maskAllImages = false;
+        await Posthog().setup(first);
+        final capturer = ScreenshotCapturer(first);
+        expect(capturer.effectiveConfig, same(first));
+
+        final second = PostHogConfig('test_project_token');
+        await Posthog().setup(second);
+        expect(capturer.effectiveConfig, same(second));
       },
     );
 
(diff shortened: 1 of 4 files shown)


posthog_flutter/test/posthog_test.dart
```diff
@@ -40,6 +42,39 @@ void main() {
...
+      'screenshot capturer resolves the live config after a second setup',
+      () async {
+        final first = PostHogConfig('test_project_token')
+          ..sessionReplayConfig.maskAllImages = false;
+        await Posthog().setup(first);
+        final capturer = ScreenshotCapturer(first);
+        expect(capturer.effectiveConfig, same(first));
+
+        final second = PostHogConfig('test_project_token');
+        await Posthog().setup(second);
```
dustinbyrne, line 78: fwiw, i wouldn't be surprised if we had more gaps with duplicate setup calls, this looks like a _really_ fringe scenario, and probably enters undefined behavior territory (unless i'm missing something)

i'm surprised you can initialize twice - imo the second should be a no-op

dustinbyrne (approved):

PR #510, opened 2026-07-31
fix(replay): keep session replay recording across a session change

turnipdabeets:
## :bulb: What this fixes

**Session replay records nothing after a session changes.** The new session never gets its meta event, so on a screen that isn't repainting the recording comes out blank. This bug is on `main` today.

Triggers: idle or maximum-duration expiry, `reset()` (logout — the most common one), `close()`/`setup()`, and `startSessionRecording(resumeCurrent: false)`.

Measured on device: **28 s of nothing → ~1 s** to the new session's first frame.

Also fixed: **an occlusion episode ending while the SDK is closed stranded capture suppressed for the life of the process**, so every later recording was blank with no channel traffic at all. Native pushes that end event once, as soon as replay goes off — i.e. right after `close()` — and Dart dropped it because the config was momentarily null. Also on `main` today.

## :hammer: How

- **Per-session state resets when the session id actually changes.** `isSessionReplayActive` became `getSessionReplayState`, returning the active flag *plus* the id, from the round trip each capture already makes. Same shape the peers use (`resetSessionStateIfNeeded` on Android, `handleSessionChanged` on iOS, `getSessionId()` on React Native).
- **The read is non-mutating on both platforms** — `peekSessionId()` on Android, the hard-wired `readOnly: true` `getSessionId()` on iOS. The send still resolves the session and so still applies the idle and maximum-duration bounds, exactly as on `main`. **Nothing about session or rollover timing changes.**
- **`reset()`, `close()` and a non-resuming start force the reset immediately**, because the platforms disagree on whether they rotate the session at all — Android keeps the id across `close()`, iOS doesn't.
- **One primitive for out-of-band sampling.** A poll tick only reaches the capture path from a post-frame callback, so a screen that never repaints needs a forced frame. Each caller needs cover for two ways a request gets swallowed: a capture already in flight, and a platform briefly not recording when the sample lands. The retry budget is cancelled **on delivery, not on capture** — otherwise a frame dropped for belonging to the previous session cancels the retries covering the one that replaced it.
- **A boundary does not force its own frame; it arms the budget.** A repaint within the next tick delivers the new session's first frame, so a host that replaces the screen after `reset()` or `close()`/`setup()` opens the new recording on the screen it replaced it with, not the one it is leaving. `start()` still forces for every other caller — app launch, a recording resumed on the same session — or a static screen would record nothing until it happened to repaint.
- **Frames carry a state generation, not just a session id.** Two kinds of frame name no session: an occlusion placeholder built before the first state read, and anything built right after a forced reset, which leaves the tracked id null until the next read. The id alone cannot separate them, so a placeholder still rasterizing when a reset landed used to ship into the recording that replaced it.

## :straight_ruler: Scope

Deliberately cut back to the bug. `setup()` and `posthog_flutter_io.dart` are byte-identical to `main`.

| Left out | Why |
|---|---|
| Android `$session_id` pre-attach | Fixes a *different* bug (a frame crossing a rotation being re-homed). It also forces the tick read onto the expiring accessor, which moves where the expiry bounds are applied — a bigger behaviour change than this fix warrants |
| Runtime `throttleDelay` | Separate feature; it made this a `minor` release |
| `setup()` rollback on failed init | Separate bug (a failed first `setup()` latches the SDK) |
| Repeated `setup()` as a no-op, and never-throw | Breaking; targets the next major |

**No public API change** (snapshot unchanged), **no breaking change**, and a single `patch` changeset.

## :warning: Known limitation

**A rotation is seen one tick late, on both platforms**, so a capture crossing an *unforced* rotation can ship one frame ahead of its meta event; the next tick repairs it. Closing this needs an id pre-attached at capture time, which in turn needs a public *expiring* session accessor on posthog-ios (`PostHogSessionManager.getSessionId(at:readOnly:)`, what its own replay integration uses) so both platforms can do it the same way.

## :green_heart: Testing

- **302 Dart tests** and **37 Kotlin tests**, analyzer, format and ktlint clean, **public API snapshot unchanged** against `main`.
- Every behavior change is mutation-tested — deleting the retry arming, the delivery-keyed cancel, the in-flight drain, a forced-reset call site, or the generation check each turns a specific named test red, and so does bumping the generation on the first id adoption.
- The retry budget is cleared only by delivery, never by `stop()`. Clearing it in `stop()` made the restarted recording depend on statement order inside `close()`, and nothing pinned either; `close()`/`setup()` with the platform reporting replay inactive for the first sample is now a test that fails if that comes back.

**Device E2E re-run at head on both platforms**, after the scope cut:

| Check | Android (Pixel 9, API 37, posthog-android 3.59.0) | iOS (iPhone 17 sim, posthog-ios 3.69.0) |
|---|---|---|
| `reset()` on a **static** screen → first frame of the new session | **1.07 s** | **1.05 s** |
| The new session's recording | **53** snapshot sends over 71 s | **184** snapshots over ~200 s |
| Steady state on a static screen | silent after the retries deliver — they cancel on delivery | 999 ms median gap once repainting |

vs **28 s** before the fix.

**Boundary content, measured back to back in one process** (only the deferral differed), with the host navigating 250 ms after the boundary:

| Boundary | Android | iOS |
|---|---|---|
| `reset()` + navigate → first frame lands | 811 ms **after** the new screen renders | 774 ms **after** |
| `close()`/`setup()` + navigate → first frame lands | 93 ms **after** the new screen renders (was 205 ms **before** it) | 59 ms **after** (was 211 ms **before** it) |
| `close()`/`setup()` on a **static** screen → first frame | 1.08 s (was 69 ms) | 1.06 s (was 78 ms) |

The instrumented state reads also show why `reset()` was never at risk: both platforms report replay **inactive for ~1 s** after it (Android `active=false` at +15 ms and +31 ms, `active=true` at +1033 ms; iOS `false` at +29 ms, `true` at +1030 ms), so a sample forced in that turn was swallowed either way. Only `close()`/`setup()`, where replay is active within ~10 ms, could ship a pre-boundary frame. The cost is confined to the last row and scales with `throttleDelay`.

## :pencil: Checklist

- [x] I updated the docs if needed. — no user-facing behaviour change to document beyond the changeset

### If releasing new changes

## 🤖 Agent context

Claude Code running the `sdks-pre-pr-review-loop` skill, plus two independent review passes over the diff and device E2E on both platforms. All claims about native behaviour are verified against the resolved versions (`posthog-android 3.59.0`, which is what the Gradle range picks — not the 3.58.0 floor — and `posthog-ios 3.69.0`).


posthog_flutter/lib/src/posthog_widget.dart
```diff
@@ -168,6 +217,16 @@ class PostHogWidgetState extends State<PostHogWidget> {
...
+    } else if (!identical(_componentsConfig, config) ||
+        _changeDetector!.interval != config.sessionReplayConfig.throttleDelay) {
+      // A close()/setup() reconfigure changed the config; components capture
+      // config-derived state (capture throttle, masking flags) at
+      // construction, so they must be rebuilt against the new one. The
+      // interval is value-compared because the same config object may be
+      // reused with throttleDelay mutated in place.
+      _changeDetector?.stop();
+      _screenshotCapturer?.cancel();
+      _initComponents(config);
```
dustinbyrne: on the surface this looks hacky: we're having to detect here if config has changed, and then reinitialize as a result - is this state not already disposed of on `close()`? imo that'd be much more straight-forward

maybe the lifecycle of this class isn't easily managed?

posthog_flutter/lib/src/posthog_widget.dart
```diff
@@ -31,6 +31,40 @@ class PostHogWidgetState extends State<PostHogWidget> {
...
+    _screenshotCapturer?.hasCapturedPlatformViews = value;
+  }
+
+  @visibleForTesting
+  PostHogConfig? get debugCapturerEffectiveConfig =>
+      // ignore: invalid_use_of_visible_for_testing_member
+      _screenshotCapturer?.effectiveConfig;
+
+  @visibleForTesting
+  void debugTriggerOnChange() => _changeDetector?.onChange();
```
dustinbyrne: this looks like code smell to me, do we _need_ all of these?

or it's a dart idiom and you should ignore me

posthog_flutter/lib/src/posthog_widget.dart
dustinbyrne: i'd considered that maybe we should just leave `_config` in place on `close()` so that the instance cannot be revived, but i suppose that's technically breaking (even though i'm not convinced it's not broken already)

posthog_flutter/lib/src/posthog_widget.dart
turnipdabeets: no we didn’t need all of those, updated!

posthog_flutter/lib/src/posthog_widget.dart
turnipdabeets: updated as well

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -409,6 +414,15 @@ class PosthogFlutterPlugin :
 
     private fun isSessionReplayActive(): Boolean = PostHog.isSessionReplayActive()
 
+    // Read on every Flutter capture tick, so it must not mutate session state:
+    // PostHog.getSessionId() runs the expiry checks and can rotate the session,
+    // peekSessionId() only reads it.
+    private fun getSessionReplayState(): Map<String, Any?> =
+        mapOf(
+            "isActive" to isSessionReplayActive(),
+            "sessionId" to PostHogSessionManager.peekSessionId()?.toString(),
```
marandaneto, line 423: blocking: Session expiry is detected one snapshot too late

`peekSessionId()` (and iOS's read-only `getSessionId()`) returns the cached ID without applying idle/max-duration expiry. The following snapshot can therefore rotate natively only after Dart has retained the old meta latch, so the new session receives a full snapshot without metadata; on a static screen there may be no later frame to repair it. Please resolve or communicate the effective session before Dart compares IDs, with Android/iOS parity.

posthog_flutter/lib/src/posthog_widget.dart
```diff
@@ -70,6 +70,19 @@ class PostHogWidgetState extends State<PostHogWidget> {
...
+  }
+
+  /// A session change Dart knows about before the capture path could observe it:
+  /// `reset()` and `close()` (both platforms end the session), and a start that
+  /// does not resume — which rotates on iOS but on Android may not rotate at
+  /// all, since `PostHog.startSessionReplay` returns early while recording is
+  /// already active. The drop is requested directly rather than waiting for a
+  /// tick to read the new id.
+  void _onForcedReplaySessionReset() {
+    _screenshotCapturer?.resetSessionStateIfNeeded(null, force: true);
```
marandaneto, line 85: blocking: Forced resets do not capture the new session’s initial frame

This clears replay state but does not request a frame. `reset()` and `startSessionRecording(resumeCurrent: false)` leave `sessionRecordingActive` true, so the detector is not restarted; a static screen can leave the new session blank indefinitely. Please schedule an immediate sample after the native boundary completes, and retry if a capture was already in flight.

posthog_flutter/lib/src/posthog.dart
```diff
@@ -59,6 +62,16 @@ class Posthog {
   /// ensure `com.posthog.posthog.AUTO_INIT: false` is set in your native
   /// configuration.
   Future<void> setup(PostHogConfig config) {
+    // Mirrors the Android and iOS SDKs, which ignore a second setup; without
+    // this guard the Dart layer would half-apply the new config while native
+    // keeps the old one.
+    if (_config != null) {
```
marandaneto, line 68: blocking: Failed setup attempts permanently latch the no-op guard

`_config` is assigned before `_posthog.setup(config)` completes, so any failure makes this guard treat the SDK as successfully initialized and future calls return success without retrying. `PosthogFlutterIO.setup()` also catches `PlatformException`, creating the same apparent-success state when native setup failed. Track pending/completed setup separately, roll back on failure, and propagate initialization errors.

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
turnipdabeets: Fixed in 77fc095 — the Android tick now reads `PostHog.getSessionId()`, the *expiring* accessor, not `peekSessionId()`, so idle and max-duration are applied on the read Dart compares against (`PosthogFlutterPlugin.kt:443-451`). `PosthogFlutterPluginTest.sessionReplayState_resolvesSessionIdBeforeIsActive` pins the read order, since resolving the id can rotate it and stop replay synchronously.

Parity is deliberately not symmetric, and the PR body says so: iOS's only public accessor is hard-wired `readOnly: true`, so pre-attaching there would trade the expiry bounds for attribution. iOS therefore keeps the bounds at send time and sees a rotation one tick late, which `onSessionRotated` repairs on the next tick. Closing that gap needs `PostHogSessionManager.getSessionId(at:readOnly:)` made public in posthog-ios — tracked in `NATIVE_BEHAVIOR.md` rows 2 and 7.

posthog_flutter/lib/src/posthog_widget.dart
turnipdabeets: Fixed — this is exactly what `_ensureSampleLands()` does, and `_onForcedReplaySessionReset` calls it (`posthog_widget.dart:87-112`). It covers both halves you named: the request is deferred past an in-flight capture (`_sampleRequestedDuringCapture`, drained in `_generateSnapshot`'s `finally`), and re-forced for up to three poll ticks until a frame is actually delivered — the cancel is keyed on delivery, not on capture, so a frame dropped for belonging to the old session cannot cancel the retries covering the one that replaced it.

Pinned by `replay_session_sampling_test.dart`: `reset()` on a static screen, `reset()` when the platform reports replay inactive for the first sample (the iOS remote-config window), a reset landing mid-capture, `startSessionRecording(resumeCurrent: false)`, and now `close()`/`setup()` with the same inactive-first-sample shape.

posthog_flutter/lib/src/posthog.dart
turnipdabeets: No longer applicable — the no-op guard this describes is gone from the PR. e67a3e6 dropped the breaking half of the `setup()` contract and ba365fc split the rollback into its own change; `setup()` is now byte-identical to `main`, so there is no `_config`-based latch here to fail open.

The underlying bug you describe is real, though, and it is why it was split out rather than dropped: a failed first `setup()` latching the SDK is a separate fix targeting its own PR, with repeated-`setup()`-as-a-no-op and never-throw going to the next major.

posthog_flutter/lib/src/replay/screenshot/screenshot_capturer.dart
```diff
@@ -76,21 +86,69 @@ class ScreenshotCapturer {
...
+  /// catch a rotation observed by the same tick that produced the frame: that
+  /// tick writes the tracked id before building it.
+  ///
+  /// A frame naming no session is always current: it carries no attribution of
+  /// its own, so the send resolves whichever session is live and it cannot land
+  /// in a dead one. Without this, a placeholder built before the first tick had
+  /// read an id would be dropped by that tick adopting one, leaving the episode
+  /// showing the uncovered Flutter tree instead of the cover.
+  bool sessionStillCurrent(String? sessionId) =>
+      sessionId == null || _replaySessionId == sessionId;
```
marandaneto: blocking: Forced resets do not invalidate unnamed frames

A placeholder captured before the first native state poll has `sessionId == null`. Because forced `reset()`/`close()`/non-resuming starts also leave `_replaySessionId` null, this condition accepts that stale in-flight frame and can send a full snapshot into the new native session without its metadata. Please attach a monotonically increasing generation to each `ImageInfo`, increment it on every forced reset, and reject earlier generations while keeping initial session-ID adoption in the current generation.

turnipdabeets: Fixed in `3005f14` — frames now carry a state generation, bumped on forced resets but not on the first id adoption, so the placeholder that has to survive that read still does.

One widening: a placeholder can't arrive bare (it hardcodes `shouldSendMetaEvent = true`) — the bare case is a normal capture whose state read returned active with a null id, which both plugins allow. The generation covers both.

posthog_flutter/lib/src/posthog.dart
```diff
@@ -364,7 +364,12 @@ class Posthog {
   /// The SDK will behave as if it has been [setup] for the first time.
   ///
   /// Returns a [Future] that completes when the reset request has been queued.
-  Future<void> reset() => _posthog.reset();
+  Future<void> reset() async {
+    // Bumped before the platform call because native rotates inside that round
+    // trip.
+    PostHogInternalEvents.requestReplaySessionReset();
```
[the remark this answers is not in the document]
turnipdabeets: `close()`/`setup()` was real — measured on device, the frame shipped ~200 ms *before* the host's replacement screen rendered; `1c9b85f` leaves a boundary's first frame to the tick budget, and it now lands after it (Android +93 ms, iOS +59 ms).

`reset()` was never at risk: both platforms report replay inactive for ~1 s afterwards, so that sample was swallowed either way.

marandaneto (approved):
Worth running tests on emulators and browser since replay changes are sensitive and might have side effects 
Lets not just trust the bots for replay changes, otherwise lgtm

turnipdabeets:
> Worth running tests on emulators and browser since replay changes are sensitive and might have side effects Lets not just trust the bots for replay changes, otherwise lgtm

Yes it has been testing on those but will do one more pass since there’s been changes.

PR #511, opened 2026-08-04
feat: add push notification support

ioannisj:

PostHog Workflows can send push notifications, and both native SDKs already ship the full feature (posthog-ios 3.69.0 via PostHog/posthog-ios#729, #735 and #743, posthog-android 3.58.0 via PostHog/posthog-android#642 and #667). Flutter apps can't reach any of it: the plugin never forwarded the push config flags and the manual APIs weren't bridged.

Ports it as a config + API bridge. The native SDKs own the whole subscription state machine (retry with backoff, durable pending records, 401 re-mint, re-registration on `identify`/`reset`, opt-out gating); nothing is reimplemented in Dart.

What's new:

- `PostHogConfig` gains `capturePushNotificationSubscriptions` and `capturePushNotificationOpened` (both default `true`, matching the natives) plus a `pushIdentityProvider` hook for projects that require identity verification.
- Three manual APIs: `registerPushNotificationToken`, `unregisterPushNotificationToken` and `capturePushNotificationOpened`. These are more load-bearing on Flutter than on native: `firebase_messaging` apps handle push in Dart, so live token refresh and warm-start taps are only reachable through them.
- The identity mint is the plugin's first native to Dart call that expects a reply. A closure can't cross the channel, so `toMap()` sends `pushIdentityProviderEnabled` and native calls `pushIdentityProvider` back over the channel when it needs a token. With multiple Flutter engines, the engine that installed a provider owns the mint route; when it detaches, a surviving provider-enabled engine gets promoted.
- AUTO_INIT apps opt out with the `com.posthog.posthog.CAPTURE_PUSH_NOTIFICATION_SUBSCRIPTIONS` / `com.posthog.posthog.CAPTURE_PUSH_NOTIFICATION_OPENED` keys (Info.plist on iOS, AndroidManifest `<meta-data>` on Android). The provider hook isn't available there, since the native SDK is set up before any Dart runs.
- Native floors move to the first releases carrying push: podspec and `Package.swift` to `3.69.0`, gradle to `[3.58.0,4.0.0)`. Web is a no-op (posthog-js has no push feature).

Platform divergences are kept exactly as the natives define them: `subtitle` is iOS-only (the Android bridge drops it), a null `appId` falls back to the bundle id on iOS and to the Firebase `projectId` (via reflection, no Firebase dependency added) on Android, and register/unregister no-op on macOS while opened capture works there.

One trap for the docs: iOS registers the raw APNs token, Android the FCM token. `FirebaseMessaging.instance.onTokenRefresh` yields FCM tokens on both platforms, so wiring it must be guarded with `Platform.isAndroid` or it overwrites the correct APNs token on iOS and silently breaks iOS delivery. The full `firebase_messaging` wiring guide is a follow-up to posthog.com docs.

Out of scope: notification rendering, channels and permission prompting stay with the host app, and no Firebase/APNs credentials were added to `example/`.

- `flutter test` (265 tests): the argument maps for the three APIs (including `appId` omitted when null), the `pushIdentityProvider` round-trip driven by an inbound `MethodCall`, a throwing provider returning `null`, and config serialization.
- `./gradlew :posthog_flutter:testDebugUnitTest`: config-flag plumbing, the Firebase `projectId` fallback, `subtitle` being dropped, and the multi-engine mint-route ownership cases (anchoring, no steal on secondary setup, promotion on detach).
- `make checkFormatDart analyzeDart formatKotlin formatSwift`; `api/posthog_flutter.api.json` regenerated and gated by `checkApiDart` in CI.
- There's no Swift test target under `darwin/`, so the iOS bridge itself is covered by review only. The wire contract (`POST /api/push_subscriptions/` body, retry, 401 re-mint) lives in the native SDKs and is covered by their suites; I didn't re-verify it end to end from Flutter.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj


posthog_flutter/lib/src/posthog_config.dart
```diff
@@ -200,6 +209,56 @@ class PostHogConfig {
...
+  /// to send without one.
+  ///
+  /// The token is minted by your backend (HS256, with `sub` = distinctId,
+  /// `app_id`, and `aud` = `posthog:push_identity`), never in the app.
+  ///
+  /// The native SDKs cache the result per `(distinctId, appId)` and give this
+  /// callback 10 seconds before falling back to an unauthenticated request, so
+  /// a slow or throwing implementation degrades rather than blocking delivery.
+  ///
+  /// On iOS this requires programmatic setup: with Info.plist auto-init
```
marandaneto: blocking: This limitation also applies to Android. Android auto-init configures the native SDK before Dart runs, and the later `Posthog.setup()` call is ignored because the native SDK is already enabled. Consequently, `pushIdentityProvider` is never installed and identity-verified push registrations fail. Please document that Android auto-init must also be disabled, or support installing the provider after auto-init.

marandaneto (approved):

example/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -243,14 +247,16 @@
...
 			);
 			buildRules = (
 			);
 			dependencies = (
 				33CC11202044C79F0003C045 /* PBXTargetDependency */,
 			);
 			name = Runner;
+			packageProductDependencies = (
+				78A3181F2AECB46A00862997 /* FlutterGeneratedPluginSwiftPackage */,
+			);
```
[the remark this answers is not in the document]
ioannisj: I don't think this holds. The package generation isn't gated on the SPM feature flag, it's gated on whether the pbxproj is already migrated. `darwin_dependency_management.dart` handles the disabled-but-migrated case explicitly and still generates `Package.swift`, just with no dependencies, and the early return in `swift_package_manager.dart` only fires when there are no dependencies *and* the project isn't migrated.

Checked it rather than reasoning about it: disabled SPM, deleted `Flutter/ephemeral/Packages` outright, rebuilt. The package regenerated as the no-dependency variant and resolved fine, nothing in the build output referencing `FlutterGeneratedPluginSwiftPackage`.

Worth noting the pbxproj hunk is Flutter's own migrator output, those object UUIDs are hardcoded in `flutter_tools`.

posthog_flutter/lib/src/posthog_config.dart
ioannisj: Yeah, you're right. Same guard on both platforms (`PostHog.kt:150` and `PostHogSDK.swift:139` both bail with "Setup called despite already being setup!"), so the config carrying the provider gets dropped either way.

Documented it for both in 1da2d2f rather than supporting install-after-auto-init. That second option is doable, both SDKs read `pushIdentityProvider` lazily on every mint so assigning it post-setup would work, but it's more surface than the gotcha seemed to warrant. Happy to go that way instead if you'd prefer.

PR #531, opened 2026-08-14
fix: normalize compliance timestamps to UTC

marandaneto:

The SDK compliance adapter accepted capture timestamp overrides with an explicit timezone offset and uploaded them unchanged. Compliance capture payloads should use UTC timestamps so the timestamp format is consistent while preserving the same instant.

This change converts valid staged timestamps to UTC before adding events to the upload batch. If a timestamp is invalid, the adapter now uses the current UTC time instead of returning a 500 response.

The unrelated web exception handling change was removed from this PR and moved to #532. Until that PR merges, this PR temporarily suppresses the new `unawaited_return_in_try_block` analyzer diagnostic for the existing web implementation. PR #532 will remove the suppression after updating from `main`.

- `cd sdk_compliance_adapter && dart analyze`
- `cd sdk_compliance_adapter && flutter test test/timestamp_test.dart`
- `make analyzeDart`
- `cd posthog_flutter && flutter pub publish --dry-run`

The tests verify both offset-to-UTC conversion and the current-time fallback for malformed timestamps through a real adapter flush. Dart analysis and package validation complete without warnings with the temporary suppression.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi updated the timestamp handling based on review feedback. The malformed timestamp behavior was explicitly chosen to fall back to the current UTC time. The unrelated published-package change was moved to #532. The analyzer suppression is temporary and must be removed by #532 after this PR merges.

### DIFF
--- a/.changeset/calm-clocks-align.md
+++ b/.changeset/calm-clocks-align.md
@@ -1,0 +1,5 @@
+---
+'posthog_flutter': patch
+---
+
+Normalize compliance adapter capture timestamp overrides to UTC before upload.
--- a/.changeset/calm-clocks-align.md
+++ b/.changeset/calm-clocks-align.md
@@ -1,5 +1,0 @@
----
-'posthog_flutter': patch
----
-
-Normalize compliance adapter capture timestamp overrides to UTC before upload.
--- a/example/analysis_options.yaml
+++ b/example/analysis_options.yaml
@@ -7,6 +7,15 @@
 
 # The following line activates a set of recommended lints for Flutter apps,
 # packages, and plugins designed to encourage good coding practices.
+analyzer:
+  exclude:
+    - build/**
+    - android/**
+    - ios/**
+    - web/**
+    - windows/**
+    - macos/**
+    - linux/**
 include: package:flutter_lints/flutter.yaml
 
 linter:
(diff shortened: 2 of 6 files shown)


arnohillen (approved):
<small>(created by claude code)</small>

[arno's agent] Verified at 2b7c9e71: DateTime.parse().toUtc() converts offset inputs to the same instant and the integration test proves conversion through a real flush; the adapter was the only unnormalized emission path. Note the compliance workflow still pins harness 0.9.0 so the new contract tests do not run here yet.

arnohillen:
<small>(created by claude code)</small>

[arno's agent] The second commit removes the changeset but also changes posthog_flutter_web.dart in the published package (the return await fix alters captureException error handling on web), so that fix will ship without a changelog entry.

arnohillen:
<small>(created by claude code)</small>

[arno's agent] DateTime.parse throws FormatException on non-ISO-8601 input, so a malformed staged timestamp now makes /capture return 500 instead of passing the value through unchanged; worth confirming the harness never stages non-ISO timestamps.

marandaneto:
Addressed both review notes:

- Reverted the unrelated web `captureException` change from this PR in dcc1db1. It now has its own changeset and browser regression test in draft PR #532.
- Invalid staged timestamps now fall back to the current UTC time instead of throwing and returning 500. This is covered by an adapter integration test in cb910f4.

PR #537, opened 2026-08-16
fix(surveys): draw icons without Material Icons font

### LINKED ISSUE #536: Survey close/check icons render as tofu in apps without the MaterialIcons font
### Bug description

The survey widgets render Material icons, which produce a missing-glyph box (tofu) in any app that does not bundle the `MaterialIcons` font:

- `lib/src/surveys/widgets/survey_bottom_sheet.dart:200` — `Icons.close` for the dismiss button
- `lib/src/surveys/widgets/survey_choice_button.dart:95` — `Icons.check` for selected choices

Bundling `MaterialIcons` is opt-in (`uses-material-design: true`), and apps built on `WidgetsApp` rather than `MaterialApp` — with their own icon font — have no reason to set it. Ours does not, so the survey's close button renders as `?` in a box.

The button still works, it just looks broken. Screenshot attached below in the thread.

### Why this is worth fixing here

The rating scale already avoids this: `lib/src/surveys/widgets/rating_icons.dart` custom-paints its emoji shapes rather than pulling them from a font. Close and check are the only two places that reach for Material. So the fix is consistent with what the package already does elsewhere, and it removes an implicit `MaterialApp` dependency from an SDK surface that otherwise composes fine into a non-Material app.

Asking every host app to ship a ~1.6MB font it does not otherwise use, purely so an SDK-rendered sheet can draw an X and a checkmark, is a heavy ask.

### Suggested fix

Draw both as paths, the same way `rating_icons.dart` does — an X and a checkmark are a handful of lines each — or ship them as a tiny bundled font asset in the package.

Failing that, documenting `uses-material-design: true` as a requirement for surveys in the Flutter surveys installation guide would at least make it diagnosable. Right now the symptom gives no hint about the cause.

### Environment

- `posthog_flutter` 5.36.2
- Flutter 3.44.5, iOS 18.3 simulator
- Host app uses `WidgetsApp.router`, no `uses-material-design` in `pubspec.yaml`

Related: #534 (nearest-navigator rendering), #535 (no imperative display API). All three come from the same place — the Flutter survey UI assumes a Material host app.

marandaneto:

Apps that use `WidgetsApp` or do not set `uses-material-design: true` do not bundle the Material Icons font. Survey close and check icons then render as missing-glyph boxes.

This change draws both icons with `CustomPainter` using the canonical Material Icons 24px path coordinates. It keeps the existing geometry, sizes, and colors without requiring a font asset. Fixes #536.

- `flutter analyze`
- `flutter test -r expanded`
- Widget tests render the close and check icons under `WidgetsApp` without a Material Icons font.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent using `openai-codex/gpt-5.6-sol` and checked with Pi autoreview. The implementation copies the canonical Apache-2.0 Material Icons path geometry into one custom painter instead of adding a bundled font asset.

### DIFF
--- a/.changeset/calm-icons-draw.md
+++ b/.changeset/calm-icons-draw.md
@@ -1,0 +1,5 @@
+---
+'posthog_flutter': patch
+---
+
+Draw survey close and check icons without requiring the Material Icons font.
--- a/posthog_flutter/lib/src/surveys/widgets/survey_bottom_sheet.dart
+++ b/posthog_flutter/lib/src/surveys/widgets/survey_bottom_sheet.dart
@@ -15,6 +15,7 @@ import 'open_text_question.dart';
 import 'rating_question.dart';
 import 'choice_question.dart';
 import 'confirmation_message.dart';
+import 'survey_icon.dart';
 
 /// A bottom sheet that displays a survey to the user.
 class SurveyBottomSheet extends StatefulWidget {
@@ -197,8 +198,8 @@ class _SurveyBottomSheetState extends State<SurveyBottomSheet> {
                     mainAxisAlignment: MainAxisAlignment.end,
                     children: [
                       IconButton(
-                        icon: Icon(
-                          Icons.close,
+                        icon: SurveyIcon(
+                          type: SurveyIconType.close,
                           color: widget.appearance.closeButtonColor,
                         ),
                         onPressed: () => _handleClose(),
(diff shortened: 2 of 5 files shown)


arnohillen (approved):
<small>(created by claude code)</small>

Verified the painter paths match the canonical Material Icons 24px close/check geometry coordinate-for-coordinate, no Icons.* usage remains anywhere in posthog_flutter/lib, and the widget tests assert the exact failure mode from #536; backwards compatible since SurveyAppearance colors are non-nullable.

PR #542, opened 2026-08-20
fix: Drop events when beforeSend callbacks throw

marandaneto:

A throwing `beforeSend` callback was logged and skipped, which allowed the original or previously modified event to continue through the remaining callbacks and reach the native capture queue. This conflicts with the fail-closed hook contract and can send an event that the callback did not approve.

This change logs the callback exception, stops the callback chain, and drops the event. Each successful callback still passes its returned event to the next callback.

- `make checkFormatDart`
- `make analyzeDart`
- `make checkApiDart`
- `make test`
- `pnpm changeset status`

The tests cover synchronous and asynchronous callback exceptions, stopping later callbacks, preventing method-channel capture, and passing a replacement event through the callback chain.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

A Pi coding agent inspected the existing implementation and tests, strengthened chain propagation coverage, ran the repository validation commands, and ran the isolated autoreview helper. The implementation returns `null` immediately after a callback throws so the event cannot continue to the platform channel.

### DIFF
--- a/.changeset/calm-hooks-close.md
+++ b/.changeset/calm-hooks-close.md
@@ -1,0 +1,5 @@
+---
+'posthog_flutter': patch
+---
+
+Drop events when a before-send callback throws.
--- a/posthog_flutter/lib/src/posthog_flutter_io.dart
+++ b/posthog_flutter/lib/src/posthog_flutter_io.dart
@@ -64,8 +64,10 @@ class PosthogFlutterIO extends PosthogFlutterPlatformInterface {
         if (result == null) return null;
         event = result;
       } catch (e) {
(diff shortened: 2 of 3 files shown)


turnipdabeets (approved):
Approving. `_runBeforeSend` is the single choke point for `capture`, `screen`, and `captureException`, and the same Dart layer serves web, so one `return null` covers every surface the spec scopes to Flutter hooks. Sync and async throws are both covered, and rewriting the ordering test to return a fresh `PostHogEvent` is a nice side effect — it now actually proves chain propagation rather than relying on in-place mutation.

PR #543, opened 2026-08-20
fix(replay): keep the native-screen cover across a session boundary

turnipdabeets:

**A session boundary landing under a native screen drops the cover.** `reset()`, or a `close()`/`setup()` pair, turns session replay off for around a second. The occlusion detector polls once a second and reads that as "recording stopped", so a tick landing inside that window ends the episode **while the native screen is still on top**:

- Dart lifts its capture suppression, resumes Flutter capture, and forces a sample — recording the Flutter tree from underneath a native screen the user is actually looking at.
- The next tick re-detects the same, unchanged cover as a **new episode**, re-handshaking the bridge for a cover that never went away.

Found while device-testing #510, and pre-existing on `main` — the branch that surfaced it is Dart-only, and this path (`PosthogFlutterPlugin.kt:968`, `PosthogFlutterPlugin.swift:888` on `main`) is untouched by it. Independent of #510 and mergeable in either order.

## :hammer: How

Hold the episode while the app is **still covered**, instead of ending it the moment replay reads inactive. Three constraints shaped it, all from review:

- **Hold for as long as the cover is up** — not on a timer. An elapsed bound can only fire while replay is inactive, i.e. while nothing is being recorded, so ending the episode there buys nothing and reopens the very window this closes: Dart lifts suppression, replay returns before the next tick re-detects the cover, and a frame of the hidden Flutter UI can ship. A bound also cannot self-heal a coverage reading that never recovers, because a stuck reading only shows up while replay is *active*, where this branch never runs — true on `main` too.
- **Coverage drives the ending, not time.** A dismissed cover ends its episode through the not-covered debounce within a tick, whatever replay is doing; detector teardown ends it explicitly.
- **Debounce the inactive branch too, per run.** A native→native handoff transiently reads not-covered and that first read can land on either side of the boundary, so the inactive branch applies the same increment-and-return the active path does — and clears the counter while covered, so a second handoff in the same episode is guarded too, not just the first.
- **Re-handshake on resume.** `enableNativeBridge` acceptance requires an active recording, so a boundary landing while Dart's enable is in flight refuses it. A hold that ends with the cover still up and no bridge re-handshakes under a new episode, reusing the cover-swap path, so there is still no end event and suppression never lapses.

The hold is a boolean: no constant, no timestamp, no elapsed arithmetic.

**Device runs on Android (Pixel 9, API 37) and iOS (26.4 simulator)**, with a temporary hook that forces the replay-inactive window and refuses a handshake mid-flight (removed before commit). Marks bracket each cover so log ordering says whether an episode ended *under* it or *at its dismissal*.

| Scenario | Result, both platforms |
|---|---|
| Boundary lands mid-cover, bridge already granted | Episode intact, ends at dismissal |
| Boundary refused mid-handshake | Placeholder episode, then a **fresh bridged episode while the cover is still up** |
| Cover dismissed 2 s into a 12 s outage | Episode ends promptly at dismissal |
| 40 s outage, cover stays up throughout | **No end at 15 s or 31 s**; the episode ends only when the cover dismisses |

The re-handshake is mutation-tested on device: deleting it leaves that episode a placeholder for the entire cover.

**Known limit, unchanged by this PR:** a native→native handoff slower than one tick still ends the episode, on `main` and here alike — the pre-existing bound of the one-tick handoff debounce.

**No unit test:** the behaviour lives in a private main-thread poll driven by the SDK's replay-active flag and private activity-lifecycle counters, with no existing Kotlin/Swift coverage of `occlusionTick`. Reaching it would mean widening that state purely for the test; the device runs above are the evidence.

## :pencil: Checklist

- [ ] I added tests to verify the changes. — see **No unit test** above; verified on device on both platforms instead
- [x] I updated the docs if needed. — no user-facing API change

### If releasing new changes

## 🤖 Agent context

Claude Code, found while running the device verification #510's reviewer asked for. The first attempt at a repro was a single `reset()` under a cover, which read as a reproduction until brackets were added to the log — with them it was clear the episode had ended at dismissal all along, and the ordering had been ambiguous. That is what motivated the forced-window hook and the explicit `--- still covered ---` marks, so the before/after claim rests on ordering that cannot be misread.

The hold was one tick initially; it went to two after the measured inactive window turned out to run close to the 1 s tick interval, leaving no margin. A pure predicate helper with unit tests was written and then dropped — the tests asserted `a && b < n`, which tests the language rather than the behaviour.

### DIFF
--- a/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
+++ b/posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
@@ -985,31 +985,29 @@ class PosthogFlutterPlugin :
     // Null activity (config-change detach) fails open — the captured Flutter tree
     // has no native pixels. Known limitation: a host recreation mid-episode never
     // re-fires onActivityResumed, ending the episode early.
-    private fun isFlutterCovered(): Boolean = activity != null && isHostActivityStopped && otherResumedCount > 0
+    private fun isFlutterOccluded(): Boolean = activity != null && isHostActivityStopped && otherResumedCount > 0
 
     @OptIn(PostHogInternalReplayApi::class)
     private fun occlusionTick() {
         if (!isSessionReplayActive()) {
             if (isOccluded || bridgeEnabled) {
-                // A session boundary — reset(), or a close()/setup() pair — turns
-                // replay off until its flags reload lands. Ending the episode
-                // there lifts Dart's capture suppression while a native screen is
-                // still on top, and the tick after re-detects that same cover as a
-                // fresh episode. Hold for as long as the cover is up: nothing is
-                // captured while replay is off, so the hold costs no frames, and
-                // the cover going away is what ends the episode either way.
-                if (isFlutterCovered()) {
+                // Replay reads inactive across a session boundary — reset(), or a
+                // close()/setup() pair — until its flags reload lands, and that is
+                // not the episode ending. Ending here would lift Dart's capture
+                // suppression under a live native cover. Unbounded on purpose:
+                // nothing is captured while replay is off, so the hold costs no
+                // frames, and the cover going away ends the episode either way.
+                if (isFlutterOccluded()) {
                     heldWhileInactive = true
-                    // Per run of not-covered reads, matching the active path,
-                    // which clears this on every tick that keeps the episode.
+                    // Re-arms the debounce budget below for the next run of
+                    // not-occluded reads. A blip seen before the hold is not
+                    // carried across it, so it cannot signal a cover swap later.
                     notOccludedTicks = 0
                     return
                 } else if (notOccludedTicks < 1) {
                     // The same end-debounce the active path applies, because a
-                    // native→native handoff's first not-covered read can land on
-                    // either side of the boundary. Without it, a tick reading the
-                    // gap between one native screen finishing and the next
-                    // resuming ends the episode with no guard at all.
+                    // native→native handoff's first not-occluded read can land on
+                    // either side of the boundary.
                     notOccludedTicks++
                     return
                 }
@@ -1027,7 +1025,7 @@ class PosthogFlutterPlugin :
         }
         val resumedFromHold = heldWhileInactive
         heldWhileInactive = false
-        val occludedNow = isFlutterCovered()
+        val occludedNow = isFlutterOccluded()
         // Debounce END only: a native→native handoff (A pauses before B resumes)
         // briefly reads not-occluded; ending the episode there would flash a
         // stale Flutter frame into the native flow.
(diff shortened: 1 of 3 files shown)


posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -964,10 +971,27 @@ class PosthogFlutterPlugin :
...
+                // replay off for around a tick. Ending the episode there lifts
+                // Dart's capture suppression while a native screen is still on
+                // top, and the tick after re-detects that same cover as a fresh
+                // episode. Hold while still covered, long enough to outlast a
+                // boundary (and a burst of them) at this tick rate. Bounded,
+                // because an episode end is the only thing that releases the
+                // suppression, so a recording that really did stop must get one.
+                if (isFlutterCovered() && inactiveTicks < INACTIVE_HOLD_TICKS) {
+                    inactiveTicks++
+                    return
```
marandaneto, line 1016: blocking: Retry an interrupted bridge handshake after replay resumes

If a session boundary starts after `onNativeOcclusionChanged` is sent but before Dart’s async `enableNativeBridge` arrives, the enable is rejected while replay is inactive. This return preserves `isOccluded`, so resuming emits no transition; `bridgeEnabled` remains false and native captures stay disabled for the rest of the cover. Please re-notify Dart or retry the handshake on resume when still occluded and unbridged, on both Android and Swift, with coverage for this timing.

turnipdabeets: Fixed in `3b9a26a` — a hold that ends with the cover still up and no bridge now re-handshakes under a new episode, reusing the cover-swap path, so there is still no end event and suppression never lapses. Verified on device on both platforms: the refused handshake is followed by a fresh bridged episode while the cover is still up, and deleting that term leaves the episode a placeholder for the whole cover.

An independent review pass also caught two more, fixed in `58b65d3`: the hold was counted in ticks, but nudges run ticks synchronously so the budget could burn in milliseconds (now elapsed time), and the inactive branch read coverage without the end-debounce the active path applies (now mirrored).

marandaneto (approved):

posthog_flutter/android/src/main/kotlin/com/posthog/flutter/PosthogFlutterPlugin.kt
```diff
@@ -964,10 +977,38 @@ class PosthogFlutterPlugin :
...
+                // there lifts Dart's capture suppression while a native screen is
+                // still on top, and the tick after re-detects that same cover as a
+                // fresh episode. Hold while still covered: nothing is captured
+                // while replay is off, so holding costs no frames.
+                if (isFlutterCovered()) {
+                    val now = SystemClock.uptimeMillis()
+                    if (inactiveHoldStartedAt == 0L) {
+                        inactiveHoldStartedAt = now
+                    }
+                    if (now - inactiveHoldStartedAt < INACTIVE_HOLD_MS) {
```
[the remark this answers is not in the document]
turnipdabeets: Fixed in `cfcf976` — the bound is deleted rather than raised, on both platforms. It could only fire while replay was inactive, so ending the episode there bought nothing and reopened exactly this window; and it could not do the job it was added for either, since a coverage reading that never recovers only shows up while replay is *active*, where this branch never runs.

The episode is now held for as long as the cover is up, and ending is driven by coverage: a dismissed cover ends it through the not-covered debounce within a tick whatever replay is doing, and teardown ends it explicitly. Device-verified on Android and iOS — a 40 s outage with the cover up produces no end at 15 s or 31 s, and the episode ends only when the cover dismisses.

PR #572, opened 2026-09-11
fix(replay): warn when PostHogWidget mounts before setup

### LINKED ISSUE #571: Session replay silently never starts if PostHogWidget mounts before setup(), with no warning
### Summary

If `PostHogWidget` mounts before `Posthog().setup()` has been called, session replay never starts for that launch — and nothing is logged. The app looks healthy, the widget is in the tree, `sessionReplay: true` is in the config, and no recording is ever produced.

### Cause

`lib/src/posthog_widget.dart`, `initState`:

```dart
final config = Posthog().config;
if (config == null) {
  return;
}
```

`Posthog().config` is null until `setup()` runs. The early return skips `_initComponents` and `_changeDetector.start()`, and also skips registering the `sessionRecordingActive` listener — so even a later `startSessionRecording()` cannot revive it. The same null check appears again at line 217.

### Why this is hard to diagnose

There is no symptom. No exception, no warning, no debug output — just an absence of recordings, which is indistinguishable from "recording is disabled in project settings", "the sample rate excluded this session", "the upload failed", or "replay is not enabled in config". All of those have to be ruled out before anyone thinks to check init ordering.

In our case the analytics SDK is initialised behind a remote-config fetch, so whether `setup()` had run before `runApp()` was genuinely non-obvious. Confirming it meant reading the SDK source to establish that `setup()` assigns `_config` synchronously before its first `await`. A single log line would have answered that in seconds.

### Request

`debugPrint` a warning on the null-config path, matching the existing convention in `setup()`:

```dart
if (config.projectToken.isEmpty) {
  debugPrint('[PostHog] projectToken must not be blank. Setup skipped.');
  return Future<void>.value();
}
```

Something like:

```dart
final config = Posthog().config;
if (config == null) {
  debugPrint(
    '[PostHog] PostHogWidget mounted before Posthog().setup() — session replay '
    'will not start for this launch. Call setup() before runApp().',
  );
  return;
}
```

This is a consistency fix rather than a new behaviour: the SDK already warns for the analogous misconfiguration one layer up.

### Environment

- `posthog_flutter` 5.39.0
- Flutter 3.44.7 (stable)

marandaneto:

Fixes #571.

When `PostHogWidget` mounts before `Posthog().setup()`, it silently skips replay initialization and listener registration. Later setup or recording commands cannot start capture for that widget instance.

This adds a warning that explains the required setup order and the option to remount the widget after setup. Because replay intent is unknown before setup, the message makes this advice conditional and says it can be ignored when replay is not intended. The warning is not emitted on web, where replay uses posthog-js. Replay initialization behavior is unchanged.

- The new regression test failed before the fix because no warning was emitted, then passed after the fix. It also verifies that rebuilding does not repeat the warning.
- All 343 VM tests passed with `flutter test --no-pub --reporter expanded`.
- All 6 Chrome tests passed for `posthog_widget_test.dart` and `posthog_widget_web_test.dart`. These cover warning suppression on web and after setup with replay enabled or disabled.
- `dart analyze .`, formatting checks, and `git diff --check` passed.
- The updated message assertion failed against the original unconditional wording and passed after the review feedback change.
- Autoreview of commit `9301682` against `origin/main` reported no actionable findings.

No emulator run was needed for this Dart-only logging change. Widget tests exercise the affected mount path with a mocked native channel.

## :pencil: Checklist

- [x] I updated the docs if needed. No documentation changes were needed.

### If releasing new changes

- [x] Added a patch changeset file manually, equivalent to `pnpm changeset`.

## 🤖 Agent context

Pi investigated the issue and implemented the change in a dedicated worktree using Git, GitHub CLI, Flutter tests, Dart analysis, and the isolated autoreview helper. The session transcript is not published.

The change follows the issue's request for a diagnostic rather than adding support for late setup. Remounting after setup restores capture within the same app launch. Review feedback prompted conditional wording because mounting the widget does not prove that replay is intended. Human review is required.

### DIFF
--- a/posthog_flutter/lib/src/posthog_widget.dart
+++ b/posthog_flutter/lib/src/posthog_widget.dart
@@ -56,6 +56,13 @@ class PostHogWidgetState extends State<PostHogWidget> {
 
     final config = Posthog().config;
     if (config == null) {
+      if (!kIsWeb) {
+        debugPrint(
+          '[PostHog] PostHogWidget mounted before Posthog().setup(). '
+          'Mobile session replay will not start for this widget instance. '
+          'Call setup() before runApp(), or remount PostHogWidget after setup().',
+        );
+      }
       return;
     }
 
--- a/posthog_flutter/lib/src/posthog_widget.dart
+++ b/posthog_flutter/lib/src/posthog_widget.dart
@@ -59,8 +59,9 @@ class PostHogWidgetState extends State<PostHogWidget> {
       if (!kIsWeb) {
         debugPrint(
           '[PostHog] PostHogWidget mounted before Posthog().setup(). '
-          'Mobile session replay will not start for this widget instance. '
-          'Call setup() before runApp(), or remount PostHogWidget after setup().',
+          'If you intend to use mobile session replay, call setup() before runApp(), '
+          'or remount PostHogWidget after setup(). '
+          'Otherwise, you can ignore this message.',
         );
       }
       return;
--- a/.changeset/posthog-widget-setup-warning.md
+++ b/.changeset/posthog-widget-setup-warning.md
@@ -1,0 +1,5 @@
+---
+"posthog_flutter": patch
+---
+
+Warn when PostHogWidget mounts before Posthog().setup() on mobile, explaining that session replay requires setup before mounting or remounting the widget after setup.
--- a/posthog_flutter/test/posthog_widget_test.dart
+++ b/posthog_flutter/test/posthog_widget_test.dart
@@ -1,0 +1,82 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:posthog_flutter/posthog_flutter.dart';
+import 'package:posthog_flutter/src/posthog_flutter_platform_interface.dart';
+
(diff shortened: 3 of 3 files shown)


posthog_flutter/lib/src/posthog_widget.dart
```diff
@@ -56,6 +56,13 @@ class PostHogWidgetState extends State<PostHogWidget> {
 
     final config = Posthog().config;
     if (config == null) {
+      if (!kIsWeb) {
+        debugPrint(
+          '[PostHog] PostHogWidget mounted before Posthog().setup(). '
+          'Mobile session replay will not start for this widget instance. '
```
marandaneto: we dont know if replay was supposed to be enabled so we cant tell if this is good or bad, opinions? night be confusing for users

marandaneto: Agreed. Updated in 9301682 to say "If you intend to use mobile session replay" and explicitly say that the message can otherwise be ignored. We cannot infer replay intent while config is null. Initialization behavior is unchanged. The updated message assertion failed against the old wording and passes now. All 343 VM tests, 6 Chrome tests, and Dart analysis passed.

ioannisj (approved):

lukas-roqqu:
Thanks for picking this up so quickly. This one cost us the most time to diagnose (#571), so having a log line is a real improvement. Any idea when it will make it into a release? Happy to test a build on our side.

Issue #568, opened 2026-09-11
Session replay: maskAllTexts does not mask text drawn by CustomPainter (RenderCustomPaint is not registered)


lukas-roqqu:
### Summary

With `maskAllTexts: true`, text drawn by a `CustomPainter` onto a `Canvas` is **not** masked and is captured in the clear in session replay screenshots.

This is not a per-widget opt-out — the config claims all text is masked, the replay looks correctly masked (every `Text` widget is a blob), and the canvas-drawn values sit in plain view alongside them. There is no signal that anything was missed.

### Cause

`lib/src/replay/element_parsers/element_parsers_const.dart` registers masking for four render object types only:

```dart
ElementParsersConst(this._factory, PostHogSessionReplayConfig? config) {
  if (config?.maskAllImages ?? true) {
    registerElementParser<RenderImage>();
  }
  if (config?.maskAllTexts ?? true) {
    registerElementParser<RenderParagraph>();
    registerElementParser<RenderTransform>();
    registerElementParser<RenderEditable>();
  }
}
```

`RenderCustomPaint` is absent, so any subtree painted through `CustomPainter.paint` is walked past. A `TextPainter.paint(canvas, offset)` call inside a painter produces no `RenderParagraph`, so nothing matches and no mask rect is emitted.

### Reproduction

```dart
class ValuePainter extends CustomPainter {
  @override
  void paint(Canvas canvas, Size size) {
    final tp = TextPainter(
      text: const TextSpan(
        text: '₦2,450,000.00',
        style: TextStyle(fontSize: 24, color: Colors.black),
      ),
      textDirection: TextDirection.ltr,
    )..layout();
    tp.paint(canvas, Offset.zero);
  }

  @override
  bool shouldRepaint(covariant CustomPainter oldDelegate) => false;
}

// In a PostHogWidget subtree, with sessionReplay: true and maskAllTexts: true:
Column(
  children: const [
    Text('₦2,450,000.00'),                                 // masked
    SizedBox(width: 200, height: 40,
      child: CustomPaint(painter: ValuePainter())),        // NOT masked
  ],
)
```

Both strings are identical. In the recording the first is masked and the second is readable.

### Expected

With `maskAllTexts: true`, canvas-drawn text is masked — or, if that is not feasible, `CustomPaint` subtrees are masked wholesale by default when either masking flag is on, since their contents cannot be introspected.

### Impact

[...]

PR #573, opened 2026-09-11
feat(replay): add opt-in custom-paint masking

marandaneto:

Related to #568.

Session replay masks ordinary text but does not mask text drawn by a `CustomPainter`. Masking every custom painter by default would also hide Flutter decorations and could cover large parts of the screen.

This PR adds `PostHogSessionReplayConfig.maskCustomPaint`, defaulting to `false`. Apps can opt in with:

```dart
config.sessionReplayConfig.maskCustomPaint = true;
```

The option works independently of `maskAllTexts` and `maskAllImages`, including when both are disabled. When enabled, it masks the full bounds of `CustomPaint` widgets with a painter or foreground painter, including their children. Painter-less containers still allow their children to be checked normally. The flag is used by both Flutter screenshot capture and the web mask-region provider.

The unmodified framework scrollbar painter is excluded when it is the only painter. This prevents desktop scrollbars from masking entire viewports while keeping child masks intact. Custom scrollbar painter subclasses and widgets with an additional background painter are still masked. Frames containing a custom painter whose bounds cannot be determined, including a zero-sized painter, are skipped instead of sent with missing masks.

Default masking behavior remains unchanged. The docs and changeset clarify that custom-painted text and images require this option or an explicit `PostHogMaskWidget`. The option remains default-off because it can also mask non-sensitive Material UI, including tab bars, checkboxes, switches, and progress indicators. Flutter's debug banner uses a full-window `CustomPaint`, so the docs recommend disabling `debugShowCheckedModeBanner` when using this option.

An internal `masksAnyContent` getter now supplies the shared flag check for native capture, web mask regions, and the web warning. It does not add another public API.

- Reproduced default over-masking before making the option independent of text and image masking.
- Reproduced the review findings before fixing them. Desktop scrollbars generated viewport-sized masks on macOS, Linux, and Windows themes. A zero-sized painter drew visible text but generated no mask, and native capture still returned that frame. Web capture returned mask regions instead of skipping the unsafe frame.
- Regression tests now cover desktop scrollbar exclusions, sensitive children inside explicit scrollbars, custom scrollbar painter subclasses, and additional background painters.
- Native capture and web provider tests verify that frames with unsized painters are skipped. All eight combinations of the three masking flags are checked for the shared getter.
- All 399 Flutter tests passed.
- All 97 focused Chrome tests passed. Two native-only capture tests are skipped on Chrome.
- All 37 focused tests passed on an Android 16 emulator. Desktop scrollbar cases use the corresponding platform themes. A temporary app used a fake PostHog platform interface and mocked replay-state calls, so no events were sent to PostHog.
- Static analysis, formatting, the public API snapshot check, and `git diff --check` passed.
- Autoreview of commit `26c6fa41be2c03a3ad6508904d33c1308feb2742` against `origin/main` reported no actionable findings.

## :pencil: Checklist

### If releasing new changes

- [x] Added a minor changeset for the new configuration option.

## 🤖 Agent context

Pi implemented this feature using file tools, Git, Flutter tests, Chrome, an Android emulator, and the autoreview helper. The work was directed by @marandaneto and done in a dedicated worktree.

The human owner requested an independent `maskCustomPaint` option defaulting to `false` to avoid changing existing replay visibility. Review feedback led to the narrow framework scrollbar exclusion. For painters with unknown bounds, capture skips the frame rather than guessing ancestor bounds, since Flutter does not clip custom painting to those bounds. Human review is required.

### DIFF
--- a/posthog_flutter/lib/src/posthog_config.dart
+++ b/posthog_flutter/lib/src/posthog_config.dart
@@ -684,12 +684,8 @@ class PostHogSessionReplayConfig {
   /// Enable masking of all text and text input fields.
   /// Default: true.
   ///
-  /// When this or [maskAllImages] is enabled, CustomPaint widgets with a
-  /// painter or foregroundPainter are masked over their entire bounds,
-  /// including their children, because canvas contents cannot be inspected.
-  /// This also masks custom-painted decorations in Flutter widgets. Disable
-  /// MaterialApp's debugShowCheckedModeBanner when testing replay: its
-  /// full-window CustomPaint otherwise masks the entire screen.
+  /// Does not mask text drawn by CustomPainter. Enable [maskCustomPaint] or
+  /// wrap sensitive custom-painted widgets in PostHogMaskWidget to mask them.
   ///
   /// With [captureNativeScreens] enabled, setting this false also unmasks text
   /// on captured native screens, including native input fields (passwords,
@@ -699,9 +695,20 @@ class PostHogSessionReplayConfig {
   /// Enable masking of all images.
   /// Default: true.
   ///
-  /// Custom-painted content is masked when this or [maskAllTexts] is enabled;
-  /// see [maskAllTexts] for details.
+  /// Does not mask images drawn by CustomPainter. Enable [maskCustomPaint] or
+  /// wrap sensitive custom-painted widgets in PostHogMaskWidget to mask them.
   var maskAllImages = true;
+
+  /// Mask the full bounds of CustomPaint widgets with a painter or
+  /// foregroundPainter, including their children.
+  /// Default: false.
+  ///
+  /// This is independent of [maskAllTexts] and [maskAllImages]. Canvas contents
+  /// cannot be inspected for individual text or images, so enabling this also
+  /// masks custom-painted decorations in Flutter widgets. Disable MaterialApp's
+  /// debugShowCheckedModeBanner when using this option: its full-window
+  /// CustomPaint otherwise masks the entire screen.
+  var maskCustomPaint = false;
 
   /// Deprecated setter that forwards assigned values to [throttleDelay].
   ///
--- a/posthog_flutter/lib/src/posthog_config.dart
+++ b/posthog_flutter/lib/src/posthog_config.dart
@@ -684,6 +684,13 @@ class PostHogSessionReplayConfig {
   /// Enable masking of all text and text input fields.
   /// Default: true.
   ///
+  /// When this or [maskAllImages] is enabled, CustomPaint widgets with a
+  /// painter or foregroundPainter are masked over their entire bounds,
+  /// including their children, because canvas contents cannot be inspected.
+  /// This also masks custom-painted decorations in Flutter widgets. Disable
+  /// MaterialApp's debugShowCheckedModeBanner when testing replay: its
+  /// full-window CustomPaint otherwise masks the entire screen.
+  ///
   /// With [captureNativeScreens] enabled, setting this false also unmasks text
   /// on captured native screens, including native input fields (passwords,
   /// card numbers) you may not have built.
@@ -691,6 +698,9 @@ class PostHogSessionReplayConfig {
 
   /// Enable masking of all images.
   /// Default: true.
+  ///
+  /// Custom-painted content is masked when this or [maskAllTexts] is enabled;
+  /// see [maskAllTexts] for details.
   var maskAllImages = true;
 
   /// Deprecated setter that forwards assigned values to [throttleDelay].
--- a/posthog_flutter/lib/src/replay/element_parsers/element_object_parser.dart
+++ b/posthog_flutter/lib/src/replay/element_parsers/element_object_parser.dart
@@ -3,6 +3,7 @@ import 'package:flutter/rendering.dart';
 import 'package:posthog_flutter/posthog_flutter.dart';
 import 'package:posthog_flutter/src/replay/element_parsers/element_data.dart';
 import 'package:posthog_flutter/src/replay/element_parsers/element_parser.dart';
+import 'package:posthog_flutter/src/replay/element_parsers/element_parsers_const.dart';
 import 'package:posthog_flutter/src/replay/mask/posthog_mask_controller.dart';
 
 class ElementObjectParser {
@@ -61,6 +62,20 @@ class ElementObjectParser {
           activeElementData.addChildren(elementData);
           return elementData;
         }
+      }
+    }
+
+    final renderObject = element.renderObject;
+    if (renderObject is RenderCustomPaint &&
+        (renderObject.painter != null ||
+            renderObject.foregroundPainter != null)) {
+      // Canvas commands cannot be inspected for sensitive text or images.
+      final parser = PostHogMaskController.instance
+          .parsers[ElementParsersConst.getRuntimeType<RenderCustomPaint>()];
+      final elementData = parser?.relate(element);
+      if (elementData != null) {
+        activeElementData.addChildren(elementData);
+        return elementData;
       }
     }
 
--- a/posthog_flutter/lib/src/replay/screenshot/screenshot_capturer.dart
+++ b/posthog_flutter/lib/src/replay/screenshot/screenshot_capturer.dart
@@ -642,8 +642,9 @@ class ScreenshotCapturer {
       );
 
       final replayConfig = effectiveConfig.sessionReplayConfig;
-      final maskAllContent =
-          replayConfig.maskAllTexts || replayConfig.maskAllImages;
+      final maskAllContent = replayConfig.maskAllTexts ||
+          replayConfig.maskAllImages ||
+          replayConfig.maskCustomPaint;
 
       ui.Image? image;
       ui.PictureRecorder? recorder;
--- a/.changeset/mask-custom-painted-content.md
+++ b/.changeset/mask-custom-painted-content.md
@@ -1,5 +1,5 @@
 ---
-"posthog_flutter": patch
+"posthog_flutter": minor
 ---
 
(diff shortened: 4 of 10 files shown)


marandaneto:
Validated commit `a440e0af45e1b07aed9b549f6996386636811f28` on an Android 16 (API 36) emulator. All **28 tests passed**.

The tests exercised the actual Flutter screenshot capture and masking path and checked the resulting replay image pixels:

- With default settings, custom-painted text remains visible while ordinary text is masked.
- With `maskCustomPaint: true`, the custom-painted area is masked.
- With `maskAllTexts` and `maskAllImages` both disabled, `maskCustomPaint: true` still masks the custom-painted area.
- After disabling `maskCustomPaint` again, the captured image is byte-identical to the default-settings image.
- Unrelated areas remain visible.

The run also covered foreground painters, flag combinations, explicit masks, painter-less containers, transforms, and the debug-banner behavior when opted in.

This used a temporary integration-test app against the PR worktree. PostHog platform calls and replay-state responses were mocked, so no events were sent to PostHog. This validates on-device rendering and screenshot masking, not ingestion or playback in the PostHog UI.

lukas-roqqu:
Thanks for this. Pulled the branch and checked it against the repro from #568 (TextPainter inside a CustomPainter): masked with `maskCustomPaint = true`, untouched by default, as documented. Opt-in makes sense given how much Material UI is built on CustomPaint. Any rough idea when this and #574 will ship? We are holding our masking approach until they land.

marandaneto:
@lukas-roqqu i just need reviews from the team and after merged the release takes minutes/max hours so today most likely

lukas-roqqu:
@ioannisj could you take a look at this one when you get a chance? It is one of the two we are waiting on before we settle our replay masking setup (details in my comment above).

ioannisj:
> @ioannisj could you take a look at this one when you get a chance? It is one of the two we are waiting on before we settle our replay masking setup (details in my comment above).

Sure, bumped this up in my PR review list for today. Will review shortly

lukas-roqqu:
> @lukas-roqqu i just need reviews from the team and after merged the release takes minutes/max hours so today most likely

Thanks man🙏

posthog_flutter/lib/src/replay/element_parsers/element_object_parser.dart
```diff
@@ -64,6 +65,20 @@ class ElementObjectParser {
       }
     }
 
+    final renderObject = element.renderObject;
+    if (renderObject is RenderCustomPaint &&
```
ioannisj, line 69: I think this masks every scroll view on desktop, and on web when the browser is on a desktop OS. `RawScrollbar` wraps the whole viewport in `CustomPaint(foregroundPainter: scrollbarPainter, child: ...)` (scrollbar.dart:2268 on 3.44.8) and `MaterialScrollBehavior.buildScrollbar` puts a `Scrollbar` around every vertical scrollable on linux/macOS/windows, so with the flag on each `ListView` body becomes a black rectangle. `TabBar` goes the same way (`CustomPaint(painter: _indicatorPainter, child: tabs)`).

I reproduced it with a widget test: `MaterialApp` + `Scaffold` + `ListView` with only `maskCustomPaint` on and the banner off. Under `TargetPlatform.macOS` the one mask rect is `(0, 56, 800, 600)`, exactly the list bounds, and the masked frame is solid black below the app bar. Same tree under `TargetPlatform.android` gives no rects at all, so it is the framework scrollbar and nothing else in MaterialApp.

Should we skip the framework scrollbar here? Something like `painter == null && foregroundPainter is ScrollbarPainter` then don't mask and keep walking the children. Probably worth naming TabBar and the checkbox/switch/progress widgets in the doc too, plus keeping that macOS `Scrollbar(child: ListView(...))` test as the regression guard.

posthog_flutter/lib/src/replay/element_parsers/element_object_parser.dart
```diff
@@ -64,6 +65,20 @@ class ElementObjectParser {
...
+        (renderObject.painter != null ||
+            renderObject.foregroundPainter != null)) {
+      // Canvas commands cannot be inspected for sensitive text or images.
+      final parser = PostHogMaskController.instance
+          .parsers[ElementParsersConst.getRuntimeType<RenderCustomPaint>()];
+      final elementData = parser?.relate(element);
```
ioannisj: I'm not sure this covers an unsized painter. `CustomPaint(painter: ...)` with no `size` and no child in a Column lays out at `Size.zero`, so `parser?.relate(element)` returns null (`buildElementData` needs `size.isValidSize`) and this falls through with no mask. `RenderCustomPaint.paint` doesn't clip, so whatever the painter drew still lands in the screenshot and the web frame unmasked.

Could we fail closed instead, e.g. when relate() returns null mask the nearest ancestor `RenderBox` with a valid size? The `painted()` helper always passes `size: Size(200, 40)`, so a test for the unsized case would be good as well.

posthog_flutter/lib/src/posthog_config.dart
```diff
@@ -684,15 +684,32 @@ class PostHogSessionReplayConfig {
...
+  /// Mask the full bounds of CustomPaint widgets with a painter or
+  /// foregroundPainter, including their children.
+  /// Default: false.
+  ///
+  /// This is independent of [maskAllTexts] and [maskAllImages]. Canvas contents
+  /// cannot be inspected for individual text or images, so enabling this also
+  /// masks custom-painted decorations in Flutter widgets. Disable MaterialApp's
+  /// debugShowCheckedModeBanner when using this option: its full-window
+  /// CustomPaint otherwise masks the entire screen.
+  var maskCustomPaint = false;
```
ioannisj, line 725: I get why opt-in, the Scrollbar issue above makes default-on a non-starter as is. But nothing in the docs or the changeset tells someone with `maskAllTexts = true` that Flutter differs from the native SDKs here. Worth a line in the `maskAllTexts` doc and the changeset, and revisiting default-on once the Scrollbar/TabBar exclusions land? Or is default-off the long-term plan?

posthog_flutter/lib/src/replay/screenshot/screenshot_capturer.dart
```diff
@@ -642,8 +642,9 @@ class ScreenshotCapturer {
       );
 
       final replayConfig = effectiveConfig.sessionReplayConfig;
-      final maskAllContent =
-          replayConfig.maskAllTexts || replayConfig.maskAllImages;
+      final maskAllContent = replayConfig.maskAllTexts ||
```
ioannisj: [suggestion] `maskAllTexts || maskAllImages || maskCustomPaint` is now hand-written in three places: here, `web_canvas_mask_provider.dart:649`, and negated in `_warnNotOptedIn` at :348. Missing the negated site next time quietly desyncs the web warning from the real mask calculation. Worth a `bool get masksAnyContent` on `PostHogSessionReplayConfig` used at all three?

posthog_flutter/lib/src/posthog_config.dart
marandaneto: overmasking is a problem so imo default off makes sense

posthog_flutter/lib/src/replay/element_parsers/element_object_parser.dart
marandaneto: Fixed in 26c6fa4. I reproduced the viewport-sized mask with macOS, Linux, and Windows themes using `flutter test test/custom_paint_masking_test.dart`. The new regressions failed before the fix and pass now, including on Chrome and the Android emulator with desktop themes.

The exemption is limited to the exact framework `ScrollbarPainter` type, with no background painter. Children are still walked. Tests also verify that sensitive children remain masked and that custom scrollbar painter subclasses or an additional background painter are not exempt. The docs now name TabBar, checkboxes, switches, and progress indicators as other UI that can be masked.

posthog_flutter/lib/src/replay/element_parsers/element_object_parser.dart
marandaneto: Fixed in 26c6fa4. The new test confirms that a zero-sized painter still draws visible text. Before the fix, the mask walk returned no masks and `captureScreenshot()` returned the unsafe frame. The regression now passes: native capture skips the frame, and the web provider returns null to skip canvas capture. Both background and foreground unsized painters are covered.

I chose to skip the frame rather than mask the nearest sized ancestor. Custom painting is not clipped to that ancestor, so its layout bounds would not reliably cover everything drawn. This only applies when `maskCustomPaint` is enabled, and the frame-skipping behavior is documented. Host, Chrome, and Android emulator validation passed.

posthog_flutter/lib/src/posthog_config.dart
marandaneto: Expanded the changeset in 26c6fa4 to explicitly warn that Flutter’s `maskAllTexts` and `maskAllImages` inspect the widget tree, not arbitrary canvas drawing. Custom-painted content still needs `maskCustomPaint` or `PostHogMaskWidget` even when those flags are enabled. It also explains the non-sensitive Material UI that can be masked. The option remains default-off, consistent with Manoel’s answer above.

posthog_flutter/lib/src/replay/screenshot/screenshot_capturer.dart
marandaneto: Addressed in 26c6fa4. All three sites now use one `masksAnyContent` getter in an internal extension on `PostHogSessionReplayConfig`, without adding another exported public API. Added tests for all eight flag combinations and for reading the getter after changing the flags. Native capture, web mask-region, warning tests, and the public API snapshot check pass.

turnipdabeets (approved):

PR #574, opened 2026-09-11
feat(replay): add selective unmasking and protect sensitive inputs

### LINKED ISSUE #569: Session replay: no unmask primitive on Flutter, and masking ignores obscureText / autofillHints
### Summary

Flutter is the only PostHog mobile SDK with no way to *unmask*. Every other SDK ships the primitive:

| SDK | Unmask primitive |
|---|---|
| iOS | `postHogNoMask()` |
| Android | `ph-no-mask` tag |
| React Native | `PostHogMaskView` |
| **Flutter** | **none** |

`PostHogMaskWidget` only ever adds masking. There is no unmask widget, no allowlist, no exclusion predicate.

### Why this matters

The [privacy docs](https://posthog.com/docs/session-replay/privacy) recommend exactly the pattern Flutter cannot express:

> Instead of selectively masking, we can selectively *unmask* fields you are sure are safe.

On Flutter the only way to make a recording legible is `maskAllTexts: false` — which unmasks **everything** globally, and then requires wrapping every sensitive widget individually. That inverts the failure mode from fail-closed to fail-open: every new screen leaks by default, and the only defence is remembering to wrap.

For a team that wants readable error messages and button labels while keeping balances and account numbers hidden, there is currently no safe way to get there.

### Requests

**1. An unmask primitive**

```dart
// mask everything globally
..sessionReplayConfig = (PostHogSessionReplayConfig()..maskAllTexts = true)

// then reveal only what is known safe
PostHogUnmask(child: Text(errorMessage))
```

This restores parity with iOS, Android and React Native, and makes the documented recommendation achievable.

**2. Honour Flutter's own sensitivity signals as a masking floor**

Masking is currently structural — it matches `RenderEditable` by type and never inspects the field's own declaration. Flutter already carries that information:

```dart
TextField(obscureText: true)                                  // password
TextField(autofillHints: [AutofillHints.creditCardNumber])
TextField(autofillHints: [AutofillHints.creditCardSecurityCode])
TextField(autofillHints: [AutofillHints.oneTimeCode])
TextField(keyboardType: TextInputType.visiblePassword)
```

Today, with `maskAllTexts: false`, **all of these render in the clear** — including password and card number fields. Nothing distinguishes them from a heading.

Treating these as always-masked regardless of `maskAllTexts` would give a floor that cannot be switched off by accident. It is zero-config and granular at the same time: apps already annotate these fields for autofill and accessibility, so the information is present in essentially every real codebase and costs the developer nothing.

Together, (1) and (2) make selective unmasking safe: mask everything, reveal deliberately, and the genuinely sensitive inputs stay masked even if someone reveals too much.

### Environment

- `posthog_flutter` 5.39.0
- Flutter 3.44.7 (stable)

marandaneto:

Fixes #569.

Add `PostHogUnmaskWidget` to reveal known-safe Flutter content without disabling replay masking globally:

```dart
PostHogUnmaskWidget(child: Text('Try again'))
```

Unmasking overrides global and explicit `PostHogMaskWidget` masks in either nesting order, matching the native SDKs' explicit mask/unmask precedence. Sensitive Flutter inputs intentionally remain masked even inside unmask wrappers and with global masking disabled.

Sensitive inputs include obscured fields; password, email, phone, name, street-address and URL keyboard types; and all 66 standard Flutter autofill hints for personal, contact, address, authentication and payment data. Coverage includes Material, Cupertino, forms and direct `EditableText`, including scaled text extending beyond dense field bounds. Unknown custom hints and unannotated inputs are not automatically classified as sensitive; annotate them appropriately or keep masking enabled without unmasking them.

For nested unmasking, only the visible rectangular region is subtracted from enclosing non-sensitive masks. Overlapping sibling masks remain intact. Unsupported relative transforms, non-rectangular/custom clips and invisible regions conservatively retain enclosing masks. Web's conservative mask bounds can cover the edges of an unmasked region. This also works with the opt-in custom-paint masking merged from `main`.

On Flutter web, mounting either wrapper inside `PostHogWidget` enables canvas masking. Both share registration and cleanup, including the single recording restart needed to block the semantics DOM. To protect frames before the first mount, declare `session_recording: { canvasCapture: { maskRegionsFn: () => null } }` in `posthog.init`. Canvas recording must be enabled separately.

Mask collection reuses sensitivity metadata from the parser. Native platform views and captured native screens are unaffected. API documentation, the public API snapshot and a minor changeset are included.

- Original regression suite reproduced 62 failures against the pre-fix code on the VM and Android emulator. The browser regression also failed before implementation.
- Forty expiration-field tests failed before the expiration-hint fix and passed afterward. Web lifecycle regressions and missing sensitivity metadata were also reproduced before fixing them.
- Nine new tests reproduced incorrect nested precedence and missing personal-field masks before this follow-up. Added retained tests for both nesting orders, all 66 standard autofill hints, sensitive keyboard types, clipping, transforms, invisible/overlapping regions, rebuilds, pixels and custom-paint compatibility.
- Merged `main` at `d2a413b8eb88a3d119fab427c9297472e2b85df5`, including custom-paint masking and release 5.42.0. Final validation: **607 VM tests and 60 Chrome tests passed**.
- **245 tests passed on the Android 16 API 36 arm64 emulator** through a temporary integration app. Coverage includes custom-paint regressions, production mask-painter pixel comparisons at 1x/2x, and typing synthetic safe text/password/card/OTP/email values, toggling password visibility and unmasking, and checking captured pixels. An earlier expanded run exceeded its command time limit; the complete rerun passed.
- Verified all 66 autofill constants exist in minimum-supported Flutter 3.27.0. Runtime tests used Flutter 3.47.3 / Dart 3.13.3.
- Static analysis, formatting, API check and `git diff --check` passed.
- `autoreview --mode branch --base origin/main` on final commit `ec9af701c77201192111885efacae1429c657c87` reported no actionable findings.
- All **28 CI checks passed** on that exact head. The initial Chrome step timed out before test output; its one failed-job rerun passed without code or timeout changes. The full CI Chrome selection also passed locally (68 tests), as did the separate WebAssembly test.

The emulator uses a fake SDK platform and the production mask painter; Chrome uses a posthog-js stub. These tests do not upload recordings or verify backend ingestion. No iOS device run was performed.

## :pencil: Checklist

### If releasing new changes

- [x] Added a changeset file.

## 🤖 Agent context

Implemented with Pi using Git, Flutter and GitHub CLI in a dedicated worktree, with reproduction-first tests and an isolated Pi autoreview pass. The agreed policy is unmask-over-explicit-mask precedence with mandatory sensitive-input protection. Human review is required.


marandaneto:
### Android emulator validation

Validated commit `9a6d0e9ef77d6171155ba19c32f5d420727c98b2` on an Android 16 arm64 emulator (API 36), using Flutter 3.47.3.

All **70 tests passed**: the 69 regression tests plus a temporary interaction test that typed synthetic values into fields, enabled “show password,” and toggled the unmask wrapper off and on.

Pixel assertions and inspection of the captured source and masked images confirmed:

- The safe error message remained visible inside `PostHogUnmaskWidget`.
- Passwords, card numbers, and OTPs remained masked. The password stayed masked in the capture even when “show password” revealed it on screen.
- Explicit masks and the private sibling outside the unmask wrapper remained masked.
- Removing the unmask wrapper masked the safe content again. Restoring it revealed only the safe content.

The regression tests also checked captured pixels at 1x and 2x resolution. The earlier before/after run reproduced 62 failures against the pre-fix masking code on the emulator.

The temporary integration app used the SDK from this PR's worktree, a fake SDK platform, and the production mask painter. This validates Flutter rendering and captured replay pixels, not recording uploads or backend ingestion. No iOS device run was performed. No further SDK changes were needed.

posthog_flutter/lib/src/replay/mask/sensitive_text_input.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+
+const _sensitiveAutofillHints = {
+  AutofillHints.password,
+  AutofillHints.newPassword,
+  AutofillHints.creditCardNumber,
+  AutofillHints.creditCardSecurityCode,
+  AutofillHints.oneTimeCode,
```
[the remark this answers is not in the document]
marandaneto: Fixed in 56fd14cba20ea185989e11d53b6bfd712b4a8ebe. Added all four expiration hints: date, day, month, and year. They are already available in Flutter 3.27.0, so the minimum supported version is unchanged.

Reproduced with `flutter test test/selective_masking_test.dart --plain-name creditCardExpiration --reporter expanded`: all 40 new tests failed because the expected masks were missing before the fix, and all 40 pass now. Coverage includes Material, Cupertino, form, and direct EditableText inputs, including inside unmask wrappers with global text masking both enabled and disabled.

Validation passed: 449 VM tests, 52 Chrome tests, 110 Android 16 emulator tests, static analysis, and the public API check. Autoreview against origin/main reported no actionable findings. Updated the API documentation and changeset as well.

lukas-roqqu:
Checked this against our real PIN and OTP inputs (pinput) with `maskAllTexts = false` and the cells stay masked, both through `obscureText` and pinput's default `oneTimeCode` autofill hint. That is exactly what #569 asked for, thank you. Keen to know when it is expected to ship so we can plan the upgrade.

lukas-roqqu:
@ioannisj could you review this one too? Verified it against our real PIN and OTP inputs (comment above). This and #573 are what we are waiting on.

posthog_flutter/lib/src/replay/element_parsers/element_object_parser.dart
```diff
@@ -3,16 +3,22 @@ import 'package:flutter/rendering.dart';
...
   ElementData? relateRenderObject(
     ElementData activeElementData,
-    Element element,
-  ) {
-    if (element.widget is PostHogMaskWidget) {
+    Element element, {
+    bool unmask = false,
+    bool sensitiveText = false,
+  }) {
+    if (element.widget is PostHogMaskWidget ||
```
ioannisj: [suggestion] A `PostHogUnmaskWidget` nested inside a `PostHogMaskWidget` never punches a hole here: the mask branch on line 20 returns before `if (unmask) return null` on line 43, and the `explicit mask wins` tests pin it.

That's the reverse of iOS and Android. `findMaskableWidgets` returns on `isNoMask()` before anything else (posthog-ios `PostHog/Replay/PostHogReplayIntegration.swift:866`), Android takes the `isUnmasked()` branch first, and the sdk-specs privacy spec says unmask wins over explicit masks. Fail-closed is the safer default so I don't mind either way, but someone porting a screen from iOS gets the opposite replay for the same nesting.

Should a nested unmask win for its subtree (a `maskSubtree` flag next to `unmaskSubtree` in `_parseAllElements`), or do we keep mask-wins and note in the spec that Flutter differs on purpose? what's your read?

posthog_flutter/lib/src/replay/mask/posthog_unmask_widget.dart
```diff
@@ -0,0 +1,29 @@
...
+/// Only wrap content known to be safe. Explicit `PostHogMaskWidget` masks and
+/// sensitive text inputs always take precedence, regardless of nesting order.
+/// This does not erase masks from ancestors or overlapping widgets, reveal
+/// native platform views, or change masking on captured native screens.
+///
+/// On Flutter web, canvas masking must already be enabled through
+/// `session_recording.canvasCapture.maskRegionsFn` in `posthog.init`, or by
+/// mounting a `PostHogMaskWidget`. This widget does not enable canvas recording
+/// or masking itself. Keep it inside `PostHogWidget` on all platforms.
+class PostHogUnmaskWidget extends StatelessWidget {
```
ioannisj: [suggestion] On web this widget doesn't opt the app into canvas masking. The mask only applies when `posthog.init` declares `maskRegionsFn` or a `PostHogMaskWidget` has mounted (`posthog_flutter/lib/src/replay/web/web_canvas_mask_provider.dart:294`), and `PostHogUnmaskWidget` is a plain `StatelessWidget` with no mount hook.

So the setup this PR recommends, `maskAllTexts = true` plus a few unmask wrappers and no mask widget anywhere, records a web app fully unmasked, including the password and card fields the new `maskAllTexts` doc says stay masked regardless. The widget doc has the web caveat, the config doc doesn't.

Worth making this stateful and calling the same `notifyMaskWidgetMounted` / `notifyMaskWidgetUnmounted` pair as `PostHogMaskWidgetState`, so mounting either wrapper opts in? At minimum the web caveat on the `maskAllTexts` doc, I think.

posthog_flutter/lib/src/replay/element_parsers/element_object_parser.dart
```diff
@@ -3,16 +3,22 @@ import 'package:flutter/rendering.dart';
...
+        isSensitiveTextInput(element.widget)) {
       final elementData = _elementParser.relate(element);
```
ioannisj, line 34: [suggestion] Small one: the widget-level branch here never sets `elementData.isSensitiveText`, only the RenderEditable branch does, so `_collectMaskWidgetElements` in `element_data.dart` has to re-run `isSensitiveTextInput(element.widget)` as a third evaluation of the same predicate to keep the field masked.

Setting the flag right after `_elementParser.relate(element)` succeeds would let `element_data.dart` collapse to `widget is PostHogMaskWidget || isSensitiveText` and drop the fallback. Worth doing here?

posthog_flutter/lib/src/replay/mask/sensitive_text_input.dart
```diff
@@ -0,0 +1,41 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+
+const _sensitiveAutofillHints = {
```
ioannisj, line 4: [suggestion] `_sensitiveAutofillHints` covers password, card and OTP hints, while iOS's `sensibleTypes` (posthog-ios `PostHog/Replay/UITextInputTraits+Util.swift:12`) also force-masks email, phone, username, name and address content types. So with `maskAllTexts = false`, or inside an unmask wrapper, `TextField(autofillHints: [AutofillHints.email])` renders in the clear on Flutter and stays masked on iOS.

posthog-js only forces `password`, so this is a parity call rather than a spec violation. Should the Flutter floor match the iOS list, or is password/card/OTP the intended scope? Either way I think the `maskAllTexts` doc should say which.

posthog_flutter/lib/src/replay/mask/posthog_unmask_widget.dart
marandaneto: Fixed in 7ffb5d7, included in the current head 0f27bd9 after merging main. PostHogUnmaskWidget now uses the same mount/unmount registration as PostHogMaskWidget, so an unmask-only app opts into web canvas masking. Updated the widget and config docs to explain the shared registration, the separate canvas-recording opt-in, and the maskRegionsFn bootstrap needed to protect frames before the first wrapper mounts.

Reproduced with `flutter test --platform chrome test/web_canvas_mask_provider_test.dart --plain-name unmask --reporter expanded`: four new lifecycle checks failed before the fix and pass now. The tests cover registration before/after provider setup, safe and sensitive mask regions, a single restart across mask/unmask mounts, and frame recovery after an out-of-tree wrapper unmounts.

After merging main, validation passed: 464 VM tests, 57 Chrome tests, and 110 Android emulator tests, plus analysis, formatting, API checks, and autoreview.

posthog_flutter/lib/src/replay/element_parsers/element_object_parser.dart
marandaneto: Addressed in 7ffb5d7, included in current head 0f27bd9. The widget-level parser now sets isSensitiveText when it creates the mask data. Rectangle collection uses that flag and no longer reruns isSensitiveTextInput.

Added a focused metadata regression: `flutter test test/element_object_parser_test.dart --plain-name "records sensitivity" --reporter expanded`. It failed before because sensitive widget-level records had a false flag, and passes now. The synthetic ElementData fixture now supplies the classification that the parser records; its original mask assertions are unchanged. All 464 VM tests, 57 Chrome tests, and 110 Android emulator tests pass after merging main.

turnipdabeets (approved):

posthog_flutter/lib/src/replay/element_parsers/element_object_parser.dart
marandaneto: Addressed in 272f3cc, now published in ec9af70. Unmask overrides explicit masks in either nesting order. A descendant unmask region is subtracted from its enclosing non-sensitive mask, while sibling masks stay intact. Sensitive inputs intentionally remain masked even inside an unmask wrapper, per the agreed policy.

The nested-unmask regression failed before the fix and passes now. Added geometry, clipping, invisible-region, sibling-mask, rebuild, captured-pixel and custom-paint integration coverage. Unsafe rectangular exclusions retain the enclosing mask; these geometry limits are documented.

After merging main: 607 VM tests, 60 Chrome tests and 245 Android emulator tests passed. Exact-head autoreview was clean. Emulator tests use the production mask painter with a fake SDK platform, not backend ingestion.

posthog_flutter/lib/src/replay/mask/sensitive_text_input.dart
marandaneto: Addressed in 272f3cc, now published in ec9af70. The agreed Flutter policy now protects all 66 standard autofill hints for personal/contact/address/authentication/payment data, plus obscured fields and password, email, phone, name, street-address and URL keyboard types. These inputs remain masked with global masking disabled and inside unmask wrappers. This mandatory floor is intentional; it is not a claim that native iOS applies the same precedence inside an explicit unmask.

Eight personal-hint widget regressions failed before the fix and pass now. Added exhaustive standard-hint tests and keyboard-type coverage, and updated the configuration docs to explain both the protected categories and the limitation for unannotated/custom inputs. All 66 constants were checked against minimum-supported Flutter 3.27.0.

Final validation: 607 VM tests, 60 Chrome tests and 245 emulator tests passed, including typing an email and verifying its captured pixels stay masked. Exact-head autoreview was clean.

marandaneto:
### Updated masking policy and emulator validation

Published in `ec9af701c77201192111885efacae1429c657c87`, after merging main at `d2a413b`.

This supersedes the earlier mask-wins behavior: unmask now overrides explicit masks in either nesting order, while sensitive inputs always remain masked. Personal/contact/address fields are included, alongside authentication and payment fields.

Final local validation passed: **607 VM tests, 60 Chrome tests, and 245 Android API 36 arm64 emulator tests**. The emulator checks include custom-paint masking, safe and sensitive input typing, password visibility toggling, unmask toggling, and original-versus-masked pixel assertions. Safe nested explicit masks are revealed; passwords, cards, OTPs, email, and independent private masks stay protected. An earlier expanded emulator run hit the command time limit; the complete rerun passed.

Static analysis, formatting, API checks, and exact-final-head autoreview passed. As before, the emulator uses a fake SDK platform and the production painter, and Chrome uses a posthog-js stub; this does not validate recording upload/backend ingestion. No iOS device run was performed. CI is running for this new head.

marandaneto:
CI is now green for `ec9af701c77201192111885efacae1429c657c87`: all 28 checks passed. The initial Chrome step timed out before reporting tests; its single failed-job rerun passed without code or timeout changes. I also ran the full CI Chrome selection locally (68 passed) and its separate WebAssembly test (passed). Both approved review threads are resolved; no open inline threads remain.

lukas-roqqu:
Merged and shipped already, thank you for turning this around so fast 🤝 The updated precedence (unmask wins over explicit masks, sensitive inputs still the floor) reads right to me, and it covers our PIN and OTP fields exactly as hoped.

PR #575, opened 2026-09-11
feat(replay): allow disabling touches without stopping screenshots

### LINKED ISSUE #570: Session replay: add captureTouches flag — masking cannot protect keypad entry because tap coordinates leak the value
### Summary

Session replay captures touch coordinates unconditionally. There is no way to disable touch capture while keeping screenshot capture, at any layer — not in `PostHogSessionReplayConfig` (Flutter), and not in the native `PostHogSessionReplayConfig` either.

On a numeric keypad this means masking provides no protection: the digits are masked, but the tap positions reconstruct the value, because a keypad layout is fixed and known.

### Why masking cannot cover this

For PIN entry, passcode entry and card number entry, the secret *is* the sequence of positions. A replay of a fully masked PIN screen still shows an ordered series of taps on a keypad whose layout is public. `PostHogMaskWidget` around the keypad changes nothing — it masks pixels, not pointer events.

So the only available mitigation is to stop recording entirely for those screens, via `stopSessionRecording()` / `startSessionRecording()`. That works, and it is what the docs recommend for sensitive screens — but it is a blunt instrument. The entire screen disappears from the recording, including the parts that are useful for diagnosing drop-off: did the user reach the PIN step, how long did they hesitate, did they navigate away.

### Request

A flag to disable touch capture while screenshots continue:

```dart
..sessionReplayConfig = (PostHogSessionReplayConfig()
  ..captureTouches = false)
```

Ideally settable at runtime, so it can be scoped to a screen:

```dart
@override
void initState() {
  super.initState();
  Posthog().config?.sessionReplayConfig.captureTouches = false;
}
```

That way a PIN screen can still be recorded — showing navigation, timing and masked layout — without the taps that would leak the PIN.

### Current behaviour, for reference

In `posthog-ios`, `PostHogReplayIntegration.swift`, touch capture is gated only on whether replay is active and whether the event is a touch — there is no configuration check:

```swift
guard event.type == .touches else { return }
guard let window = UIApplication.getCurrentWindow() else { return }
guard let touches = event.touches(for: window) else { return }
```

`PostHogSessionReplayConfig` exposes `maskAllTextInputs`, `maskAllImages`, `maskAllSandboxedViews`, `maskPhotoLibraryImages`, `captureNetworkTelemetry`, `screenshotMode`, `throttleDelay`, `captureLogs` and `sampleRate` — nothing for touches.

### Environment

- `posthog_flutter` 5.39.0, PostHog iOS pod 3.71.6
- Flutter 3.44.7 (stable)

marandaneto:

Closes #570.

Masking a PIN keypad hides its pixels but not its replay touch coordinates. A known keypad layout can therefore reveal the entered value.

Add `sessionReplayConfig.captureTouches`, enabled by default, to disable touch recording on Android and iOS while masked screenshots continue. Set it before `Posthog().setup(config)`. Both native bridges forward the setting during initialization. There is no runtime toggle, and changing the Dart field after setup does not update native state.

The native changes are merged and released in [iOS 3.74.0](https://github.com/PostHog/posthog-ios/releases/tag/3.74.0) and [Android 3.64.0](https://github.com/PostHog/posthog-android/releases/tag/android-v3.64.0). These are the minimum versions required by this PR. Native implementation PRs: https://github.com/PostHog/posthog-ios/pull/823 and https://github.com/PostHog/posthog-android/pull/780.

- Merged the latest main branch without conflicts. All 351 Flutter tests and all 36 Android bridge tests pass.
- Disabled local native SDK overrides and verified resolution to published Android 3.64.0 and iOS 3.74.0 artifacts.
- Built the Android release APK and web example. Built the iOS simulator and macOS examples with both CocoaPods and Swift Package Manager.
- During implementation, separate Android emulator startups with touch capture enabled and disabled each received three keypad taps. Enabled capture sent six touch events and disabled capture sent zero. Each run sent four masked screenshots and kept replay active. This earlier runtime validation used local native builds of the changes that are now released.
- Dart analysis, formatting, API snapshot checks, and committed-branch autoreview pass.

## :pencil: Checklist

### If releasing new changes

- [x] Added a minor changeset file.

## 🤖 Agent context

Pi implemented and tested the Flutter and native changes in separate worktrees using Flutter, Gradle, adb, XcodeBuildMCP, a local test server, and autoreview. The implementation follows the requested initialization-only configuration. The session transcript remains local. Human review is required.

### DIFF
--- a/.changeset/replay-capture-touches.md
+++ b/.changeset/replay-capture-touches.md
@@ -1,0 +1,5 @@
+---
+"posthog_flutter": minor
+---
+
+Add `sessionReplayConfig.captureTouches` and an awaitable `Posthog().setCaptureTouches(bool)` on Android and iOS to disable replay touch coordinates while keeping masked screenshots. This protects sensitive keypad entry whose values can be reconstructed from tap positions even when pixels are masked. Requires the native touch-capture controls in posthog-android 3.64.0 and posthog-ios 3.74.0.
--- a/.changeset/replay-capture-touches.md
+++ b/.changeset/replay-capture-touches.md
@@ -2,4 +2,4 @@
 "posthog_flutter": minor
 ---
 
-Add `sessionReplayConfig.captureTouches` and an awaitable `Posthog().setCaptureTouches(bool)` on Android and iOS to disable replay touch coordinates while keeping masked screenshots. This protects sensitive keypad entry whose values can be reconstructed from tap positions even when pixels are masked. Requires the native touch-capture controls in posthog-android 3.64.0 and posthog-ios 3.74.0.
+Add `sessionReplayConfig.captureTouches` to disable replay touch coordinates during SDK initialization on Android and iOS while keeping masked screenshots. This protects keypad entry whose values can be reconstructed from tap positions. Requires posthog-android 3.64.0 and posthog-ios 3.74.0. Runtime changes are not supported.
--- a/api/posthog_flutter.api.json
+++ b/api/posthog_flutter.api.json
@@ -178,28 +178,6 @@
                         "isDeprecated": false,
                         "isExperimental": false,
                         "isStatic": false,
-                        "name": "setCaptureTouches",
-                        "parameters": [
-                            {
-                                "isDeprecated": false,
-                                "isExperimental": false,
-                                "isNamed": false,
-                                "isRequired": true,
-                                "name": "enabled",
-                                "relativePath": "lib/src/posthog_flutter_platform_interface.dart",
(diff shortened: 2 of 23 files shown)


marandaneto:
TODO: Add the same initialization-only `captureTouches` option to React Native after the native SDK changes are released. Default it to `true` and forward it during Android/iOS SDK setup so apps can disable replay touch coordinates while keeping screenshots. No runtime toggle.

turnipdabeets (approved):
iOS and android need to release first, but approving to unblock.

PR #581, opened 2026-09-15
feat(replay): add textMaskPolicy for glyph-level text masking

### LINKED ISSUE #580: Session replay: mask part of a text node (e.g. only the digits) instead of the whole node
### Problem Statement

Session replay text masking in Flutter is all-or-nothing per text node. `maskAllTexts` masks every `Text`/`RichText`/input, and `PostHogMaskWidget` / `PostHogUnmaskWidget` move whole subtrees between masked and clear. There is no way to say "mask the digits in this string and leave the words".

For a fintech app that is the case that matters. A balance card renders `Total balance ₦2,450,000.00`, a transaction row renders `Ikeja Electric · -₦12,500.00`, a note field holds `Rent for flat 402`. What we want in the replay is the label readable and the number hidden. Today the choices are:

- `maskAllTexts = true` (the default): the whole screen is black boxes. You can see layout and taps, but not which screen the user was on, which button they pressed, or what a row said. Product and support get almost nothing out of the replay.
- `maskAllTexts = false` and wrap every sensitive value in `PostHogMaskWidget`: every amount, account number, card number and reference in the app has to be found and wrapped, in every screen, forever. A single missed `Text` leaks a balance. And it can't be done at all for strings that mix label and value in one node (`'Account $accountNumber'`, `Text.rich` amounts with a currency span, interpolated copy from a backend).

The same applies to text inputs: a non-sensitive `TextField` holding `Room 402` is either fully masked or fully visible.

Related but not the same: #568 / #569 (defaults and unmask gaps), #574 (keypads / sensitive inputs). Those are about which nodes get masked. This is about masking part of a node.

### Solution Brainstorm

Let the app decide per text node which character ranges get masked, and have the SDK turn those ranges into rects using the glyph boxes the render objects already expose (`RenderParagraph.getBoxesForSelection`, `RenderEditable.getBoxesForSelection`). The masking pipeline already produces a list of rects per element, so partial rects slot into `ImageMaskPainter` unchanged.

Sketch of an API on `PostHogSessionReplayConfig`:

```dart
config.sessionReplayConfig.textMaskPolicy = PostHogTextMaskPolicies.digits();
```

where a policy is `PostHogTextMask Function(String text)` returning one of `all()`, `none()`, `only(ranges)`, `except(ranges)`, with presets for the common cases (`digits()`, `redact(RegExp)`, `reveal(RegExp)`). Precedence stays as it is today: `PostHogUnmaskWidget` and `PostHogMaskWidget` win over the policy, sensitive inputs (obscured / password / OTP / email / phone hints) stay fully masked, and the policy only decides for the remaining `Text`, `RichText` and non-sensitive inputs. If a policy throws or returns a range outside the text, mask the whole node rather than leak.

I have this working on a branch with tests and an example screen and will open a PR.

lukas-roqqu:

Fixes #580.

Text masking in Flutter replay is all-or-nothing per node: `maskAllTexts` blacks out every text node, and `PostHogMaskWidget` / `PostHogUnmaskWidget` move whole subtrees. Apps that render a label and a sensitive value in the same node (`'Account $number'`, a `Text.rich` amount with a currency span, a note field holding `Rent for flat 402`) can't keep the label and hide the value. With the default on, a replay of a fintech app is a screen of black boxes; with it off, every value in the app has to be found and wrapped, forever.

This adds `sessionReplayConfig.textMaskPolicy`: a function that sees each text node's rendered string and returns what to mask.

```dart
config.sessionReplayConfig.textMaskPolicy = PostHogTextMaskPolicies.digits();
```

```dart
typedef PostHogTextMaskPolicy = PostHogTextMask Function(String text);

sealed class PostHogTextMask {
  const factory PostHogTextMask.all();
  const factory PostHogTextMask.none();
  const factory PostHogTextMask.only(Iterable<TextRange> ranges);
  const factory PostHogTextMask.except(Iterable<TextRange> ranges);
}

abstract final class PostHogTextMaskPolicies {
  static PostHogTextMaskPolicy digits();               // mask runs of digits
  static PostHogTextMaskPolicy redact(RegExp pattern); // mask every match
  static PostHogTextMaskPolicy reveal(RegExp pattern); // mask everything but the matches
}
```

A custom policy is just a closure, so an app can combine patterns or decide per node:

```dart
config.sessionReplayConfig.textMaskPolicy = (text) => PostHogTextMask.only([
  for (final m in email.allMatches(text)) TextRange(start: m.start, end: m.end),
  for (final m in digits.allMatches(text)) TextRange(start: m.start, end: m.end),
]);
```

### How it works

- `TextMaskPolicyParser` runs on the render object element of a text node (`RenderParagraph` for `Text` / `RichText`, `RenderEditable` for inputs). It hands the policy `toPlainText(includeSemanticsLabels: false)` so offsets line up with the laid-out text, and turns the returned ranges into rects with `getBoxesForSelection`. A range that wraps yields one rect per line. Boxes use the full line height (tight glyph boxes left 1px slivers above and below revealed words in `except` mode).
- In `ElementObjectParser.relateRenderObject`, the existing `Text`-widget branch steps aside when a policy is set, and the `RenderParagraph` / `RenderEditable` branch asks the parser first. Everything else in the walk is unchanged.
- Precedence is unchanged: `PostHogUnmaskWidget` → `PostHogMaskWidget` and the sensitive-input floor (obscured / password / OTP / email / phone) → the policy → `maskAllTexts`. With a policy set, it decides for the text nodes it handles, so `maskAllTexts` can stay at its default and the policy still keeps labels readable. Images and custom paint are not touched.
- `masksAnyContent` includes the policy, so setting one with every `maskAll*` flag off still runs the mask walk.
- Fails closed: a policy that throws, or returns a range outside the text, masks the whole node. The reason is logged in debug.
- Flutter widget tree only, like the rest of the masking; it can be set at runtime.

### Demo

Real replay frames: each image below is the `wireframes[0].base64` payload the SDK uploaded in its `$snapshot` event, captured from the example app on an iPhone 17 Pro simulator against a local ingest endpoint. Nothing is drawn on top afterwards; this is what the replay player receives. The screen is the new **Text Mask Policy (Replay)** page in the example app.

| On screen | `maskAllTexts` (default, no policy) | `PostHogTextMaskPolicies.digits()` |
| --- | --- | --- |
| screen | off | digits |

| `reveal(RegExp(r'\b(Total balance\|VISA\|EXP\|Recent)\b'))` | custom: emails + digits | `digits()` on text inputs |
| --- | --- | --- |
| reveal | custom | inputs |

Things to look at: the balance `RichText` masks `2,450,000` and `.00` as two rects around the `₦` span; `Ada Lovelace · Account 0123456789` keeps the name and masks the number; the transaction dates mask `12` and `2026` and keep `Sep`; the note field masks `402` and keeps `Rent for flat`; the email field stays fully masked under every policy because it is a sensitive input.

- New `test/text_mask_policy_test.dart` (21 tests):
  - presets: `digits()` on an amount, a card number, separate numbers, text without digits; `redact()` on emails; `reveal()` on a currency code.
  - in the widget tree, asserting on the rects `PostHogMaskController.getMaskElements` returns against `getBoxesForSelection`: digits masks only the amount of a `Text`; `RichText` spans; offsets stay aligned when a span has a `semanticsLabel`; a non-sensitive `TextField` (`Room 402`) masks only `402`; an obscured `TextField` is still masked in full under a policy that would reveal it; `PostHogUnmaskWidget` reveals a node the policy would mask; `PostHogMaskWidget` masks a node the policy would reveal; the policy overrides `maskAllTexts = true`; `all` masks the node exactly once; `except` masks everything but the revealed range, checked by painting the masks with `ImageMaskPainter` and reading pixels; a throwing policy masks the whole node; an out-of-range range masks the whole node; a wrapped range yields one rect per line; a policy alone turns the mask walk on with every flag off; a policy set at runtime applies on the next walk.
- Full `flutter test` in `posthog_flutter`: 628 passed. `dart analyze .` clean, `dart format --set-exit-if-changed ./` clean, `make checkApiDart` up to date (snapshot regenerated with `make updateApiDart`).
- Example app on an iPhone 17 Pro simulator (iOS 26.5): the frames above, switching policies at runtime with the segmented control. Not run on Android or web.

While writing the tree tests I noticed `getMaskElements(includeAllWidgets: true)` returns the `PostHogMaskWidget` / sensitive-input rects twice (once from `extractMaskWidgetRects`, once from `extractRects`). It's pre-existing, harmless for painting, and not changed here.

## :pencil: Checklist

- [ ] I updated the docs if needed. (docs PR on posthog.com to follow once the API shape is agreed)

### If releasing new changes

- [x] Ran `pnpm changeset` to generate a changeset file (`.changeset/text-mask-policy.md`, minor)

## 🤖 Agent context

Directed by @lukas-roqqu (I can't set the assignee from a fork; please assign me). Written with Claude Code (Claude Fable 5.1), using the Dart MCP server for hot restart and glint for driving the simulator during the captures. The tests, checks and captures above were run, not described.

Decisions along the way:

- A policy function returning a sealed decision rather than a bare `RegExp` config field: a single regex can't express `reveal`/`except` or per-node choices, and this way the SDK never interprets patterns itself. The presets cover the regex cases.
- The policy decides for text nodes even when `maskAllTexts` is true, so the safe default stays on and the policy only relaxes it where it says so. Sensitive inputs and `PostHogMaskWidget` still win.
- Fail closed on a throwing policy or a bad range. Masking too much is recoverable; a leaked balance isn't.
- Range boxes use the full line height. The first captures used tight glyph boxes and `except` left thin strips of the line above and below each revealed word. `RenderEditable.getBoxesForSelection` has no `boxHeightStyle`, so its boxes are grown to the caret rect of their line instead.
- Rejected: doing this in `ImageMaskPainter` (no access to text layout) and a per-widget wrapper (doesn't help interpolated strings, and is the status quo).


posthog_flutter/lib/src/replay/element_parsers/text_mask_policy_parser.dart
```diff
@@ -0,0 +1,162 @@
...
+          range.isCollapsed) {
+        printIfDebug(
+          '[PostHog] textMaskPolicy returned a range outside the text, '
+          'masking the whole node instead.',
+        );
+        return [whole];
+      }
+      final List<TextBox> lineBoxes;
+      try {
+        lineBoxes = _boxesForRange(renderObject, range);
```
marandaneto, line 153: blocking: Partial grapheme selections can leave digits visible — For `Code 1\u20e3`, `digits()` selects the digit but not its combining mark. Flutter returns no selection box for that partial glyph, so `_rangeRects` emits no mask and exposes content the policy requested to redact. Expand masking selections to complete glyphs or fail closed when they cannot be represented safely. Reproduction: reproduced — `flutter test test/text_mask_policy_test.dart --plain-name 'review:'` with the added combining-digit regression test fails the glyph-coverage assertion on this head; a temporary whole-node fallback for empty selection boxes makes it pass.

posthog_flutter/lib/src/replay/element_parsers/element_object_parser.dart
```diff
@@ -113,6 +118,17 @@ class ElementObjectParser {
...
     if (element.renderObject is RenderParagraph ||
         element.renderObject is RenderEditable) {
+      if (textMaskPolicy != null) {
+        final masks = _textMaskPolicyParser.relate(element, textMaskPolicy);
+        if (masks != null) {
+          for (final mask in masks) {
+            activeElementData.addChildren(mask);
+          }
+          // A text node has no maskable descendants; keep the current parent.
+          return null;
```
marandaneto: blocking: Preserve unmask precedence inside WidgetSpan — A `RenderParagraph` can have widget descendants through `WidgetSpan`. Returning `null` makes their unmask markers siblings of the policy masks rather than descendants, so `subtractUnmaskRects` never applies them. Consequently, `.all()` masks a nested `PostHogUnmaskWidget` despite its documented precedence. Preserve the ancestry needed to subtract those unmask regions. Reproduction: reproduced — `flutter test test/text_mask_policy_test.dart --plain-name 'review:'` with the added WidgetSpan regression test shows the explicitly unmasked text remains covered; temporarily returning the single mask as the parent makes it pass.

marandaneto (commented):
Automated advisory code review. Two reproduced findings.

marandaneto:
i think we should adopt similar controls to https://posthog.com/docs/session-replay/privacy
eg `maskInputOptions``maskInputFn`, `maskTextSelector`, `maskTextFn` and similar controls so its not very different than the rest
eg we give a function that has the widget and you decide what to do with it, its more flexible than the current approach i think, wdyt?

marandaneto:
thanks for the pr @lukas-roqqu 
left a comment and a suggestion
moving to draft until we address those

lukas-roqqu:
Thanks for the review, both blocking findings confirmed and fixed in 127fa9e, with a regression test for each:

- The grapheme-cluster case: an empty box list for a validated range is now treated as a layout failure (same fail-closed path as everything else) instead of silently masking nothing.
- The `WidgetSpan` case: when a decision produces exactly one mask rect, the walk now keeps descending into it instead of discarding ancestry, so a nested `PostHogMaskWidget`/`PostHogUnmaskWidget` attaches as its descendant again and `subtractUnmaskRects` sees it. (A decision with more than one rect can't have this problem — a `WidgetSpan` occupies its own inline slot and can't overlap a sibling text glyph run.)

On the API shape: agreed that handing the callback only the plain text is the less flexible half of this, and I don't want to reinvent something unlike the rest of the SDK's controls. Concretely, what if `PostHogTextMaskPolicy` became `PostHogTextMask Function(String text, Widget widget)` — same callback shape you're describing (you get the widget, you decide), just typed as a decision (`all()`/`none()`/`only(ranges)`/`except(ranges)`) rather than a boolean or a masked-text string? That keeps the part that's actually load-bearing here: glyph-range precision needs *ranges*, not a yes/no or a replacement string, and the sealed return type is what makes "fails closed on a bad range" a compile-time-checked exhaustive switch rather than a convention. Everything else — precedence, presets, fail-closed behavior — stays as is; it's a one-parameter addition plus updating the presets to ignore it. Happy to push that if it lines up with what you had in mind, or if you had a different shape for the callback's return value itself, let me know and I'll match it.

lukas-roqqu:
Pushed the widened API in 2f66bd9: `PostHogTextMaskPolicy` is now `PostHogTextMask Function(String text, Widget widget)`.

One thing worth flagging before you look: the widget you get is the `RichText`/`EditableText` that actually produced the render object, not the outer `Text`/`TextField` an app writes — `Text` is sugar over `RichText`, and there's no reliable way to walk back to it across Flutter's internal composition. I updated the docs to say so plainly rather than show a `.key`-based example that would look like it works and then not. Its style, its `InlineSpan`, and (for inputs) `obscureText`/`readOnly` are all real and reachable, though.

New test proves it's actually consulted, not just accepted and ignored: two `Text` nodes with the identical string, told apart only by font weight.

Same checks as before: 631 tests pass, `dart analyze`/`dart format`/`make checkApiDart` clean.

lukas-roqqu:
Ran a deeper review pass on the last two commits before considering this settled, and it found three real ones — pushed in 3554f93.

- **The WidgetSpan fix from the first round had a gap**: it only covered decisions that collapse to one rect. `except` almost never does (its rects are the complement of the whole node), so `reveal()` on a paragraph with a nested `PostHogUnmaskWidget` in a `WidgetSpan` was still masking it — the exact bug from the first round, just not fully closed. Reproduced the leak, then confirmed it's gone.
- **A caller's ranges can throw lazily** (the shipped presets return a `.map()`), and that specific path wasn't inside either existing try/catch — it escaped all the way up and dropped the whole replay frame instead of just failing that one node closed.
- **The widget passed for a text input was `_Editable`** — Flutter's private class that creates `RenderEditable`, not `EditableText`. Nothing on it was reachable, so the doc's own example (`obscureText`, `readOnly`) never actually worked. Now walks up to the real `EditableText`, confirmed reachable, confirmed those properties come through correctly.

Two lower-severity things also came up that I'm leaving alone for now, flagging rather than silently dropping: an ellipsized/truncated text node degrades to whole-node masking (safe, just not partial — happens because there's no box for the clipped-off characters), and a policy whose pattern can match an empty string (e.g. a regex with a trailing `*`) currently masks the whole node rather than skipping the empty match — also safe, arguably more surprising than necessary, one line to change if you'd rather it didn't.

634 tests pass, checks clean, no public API change.

lukas-roqqu:
@marandaneto ready for review. The API is the shape you asked for now — `PostHogTextMask Function(String text, Widget widget)`, so the callback gets the widget and decides.

Since you moved it to draft: both your findings are fixed, and a later review pass turned up three more that are fixed too. All threads resolved, CI green.

One thing to know going in — the widget handed to the callback is the `RichText`/`EditableText` that produced the render object, not the outer `Text`/`TextField` an app writes. `Text` is sugar over `RichText` and there's no reliable way to walk back to it. The docs say that plainly rather than showing a `.key`-based example that looks like it works and then doesn't.

I'm away from 28 Sep for two weeks, so I'd rather catch any changes this week if you get a chance.

posthog_flutter/lib/src/replay/element_parsers/text_mask_policy_parser.dart
```diff
@@ -0,0 +1,237 @@
...
+      } catch (e) {
+        printIfDebug(
+          '[PostHog] textMaskPolicy range failed to lay out, '
+          'masking the whole node: $e',
+        );
+        return [whole];
+      }
+      if (lineBoxes.isEmpty) {
```
marandaneto, line 206: blocking: Nonempty selections can still omit a trailing combined digit — The new empty-box fallback fixes the earlier single-digit case, but `digits()` on `Code 12\u20E3` selects `12` and Flutter returns a box only for `1`; the combined `2⃣` remains visible because the selection is nonempty. Check grapheme boundaries or otherwise fail closed for partially representable ranges, not just entirely empty selections. Reproduction: reproduced — `flutter test test/text_mask_policy_test.dart --plain-name 'review:'` with an added `Code 12\u20E3` regression fails the final combined-digit coverage assertion on this head; the same widget and coverage assertion pass with merge-base full-text masking.

posthog_flutter/lib/src/replay/element_parsers/text_mask_policy_parser.dart
```diff
@@ -0,0 +1,237 @@
...
+    return parts;
+  }
+
+  // Full line height rather than tight glyph bounds, so an `except` decision
+  // leaves no slivers of the line above and below the revealed glyphs.
+  List<TextBox> _boxesForRange(RenderBox renderObject, TextRange range) {
+    final selection =
+        TextSelection(baseOffset: range.start, extentOffset: range.end);
+    if (renderObject is RenderParagraph) {
+      return renderObject.getBoxesForSelection(
```
marandaneto, line 248: blocking: Glyph masks leave text shadows readable — Selection boxes exclude `TextStyle.shadows`, so masking digits can leave a readable shadow copy elsewhere inside the text node. A `Code 123` fixture with a green shadow offset by `Offset(100, 0)` retains the digit's shadow after `ImageMaskPainter` runs, whereas merge-base full-node masking hides it. Account for painted effects or conservatively mask the node when glyph bounds cannot cover them. Reproduction: reproduced — `flutter test test/text_mask_policy_test.dart --plain-name 'review: digit shadows'` with the added pixel regression observes green shadow pixels instead of masked black on this head; the same widget and pixel assertions pass with merge-base full-text masking.

marandaneto (commented):
Two blocking findings from the follow-up masking review. Existing tests pass; both additional regression tests reproduce on the reviewed head.

posthog_flutter/lib/src/posthog_config.dart
```diff
@@ -741,6 +742,56 @@ class PostHogSessionReplayConfig {
...
+  ///
+  /// Precedence, highest first: sensitive inputs (`obscureText`, password,
+  /// card, and other sensitive autofill hints) are always fully masked;
+  /// `PostHogMaskWidget` masks and `PostHogUnmaskWidget` reveals its subtree
+  /// without consulting the policy; then the policy decides; [maskAllTexts]
+  /// only applies to text nodes when no policy is set.
+  ///
+  /// The policy runs on the UI thread for every captured text node; keep it
+  /// fast. It fails closed: throwing, or returning a range outside the text,
+  /// masks the whole node. It applies to the Flutter widget tree only — text
```
marandaneto: suggestion: Document the WidgetSpan fallback in the public textMaskPolicy docs — Keeping the conservative fallback for now is fine; no behavior change is requested here. Please note that a paragraph containing a WidgetSpan falls back to a whole-node mask whenever the policy produces zero or multiple rectangles, including `.none()` and `.only([])`, while nested explicit unmask regions are still respected. The rationale is explained in the parser, but this customer-visible limitation is absent from these API docs. A more precise implementation can be deferred. Reproduction: not reproduced — this documentation request is based on the existing `hasWidgetSpan && rects.length != 1` branch; no additional runtime test was run for this request.

marandaneto (commented):
Non-blocking documentation follow-up; the conservative WidgetSpan fallback can remain for now.

marandaneto:
@lukas-roqqu nice, thanks
its in the right direction, left 3 more comments otherwise lgtm

marandaneto:
> @marandaneto ready for review. The API is the shape you asked for now — `PostHogTextMask Function(String text, Widget widget)`, so the callback gets the widget and decides.
> 
> Since you moved it to draft: both your findings are fixed, and a later review pass turned up three more that are fixed too. All threads resolved, CI green.
> 
> One thing to know going in — the widget handed to the callback is the `RichText`/`EditableText` that produced the render object, not the outer `Text`/`TextField` an app writes. `Text` is sugar over `RichText` and there's no reliable way to walk back to it. The docs say that plainly rather than showing a `.key`-based example that looks like it works and then doesn't.
> 
> I'm away from 28 Sep for two weeks, so I'd rather catch any changes this week if you get a chance.

just did a review, if you dont manage to get it done bfore pto, i can push to your pr or cherry pick your changes in a new pr, no worries, thanks for the pr btw, really appreciate it

posthog_flutter/lib/src/replay/element_parsers/text_mask_policy_parser.dart
lukas-roqqu: Confirmed and fixed in a8f45fa. Reproduced it first: `digits()` selects `12`, Flutter hands back one box covering only the `1`, and the keycap `2` sits unmasked right next to it.

The range's ends are now checked against grapheme boundaries, so a range that cuts into a cluster masks the whole node instead. Regression test added — it fails on the previous head and passes here.

posthog_flutter/lib/src/replay/element_parsers/text_mask_policy_parser.dart
lukas-roqqu: Confirmed and fixed in a8f45fa. I painted your fixture through `ImageMaskPainter` and sampled the pixel 100px right of the digits: pure green, so the shadow really was a readable copy of the masked text.

A node whose style paints `shadows` now masks whole rather than per-glyph. I went with the conservative route you offered instead of growing the rects, because that wouldn't help `except` — there the escaping shadow belongs to text outside the revealed window. Pixel regression test added.

posthog_flutter/lib/src/posthog_config.dart
lukas-roqqu: Done in a8f45fa. The `textMaskPolicy` docs now list every case that falls back to a whole-node mask, the `WidgetSpan` one included, with `PostHogTextMask.none()` and `.only([])` called out and a note that a nested `PostHogUnmaskWidget` still reveals its own subtree.

No behaviour change, as you said. Two more entries joined the list from the blocking comments above.

lukas-roqqu:
@marandaneto all three addressed in a8f45fa — both blocking ones reproduced first, then fixed, with a regression test each that fails on the previous head.

The grapheme one was as you described: a range ending mid-cluster gets a box for the part before the cut and nothing after it, so the box list is non-empty and the old check sailed past it. Range ends are now checked against grapheme boundaries.

For the shadows I took the conservative option you offered rather than growing the rects, since that wouldn't help `except` — the escaping shadow there belongs to text outside the revealed window, so there's no bound to grow by.

636 tests green, analyze clean.

posthog_flutter/lib/src/replay/element_parsers/text_mask_policy_parser.dart
marandaneto: Pushed a follow-up fix in 6b1ce5b612e9b101ac0b5728d979389b4375f6ae for inherited shadows. Flutter’s TextSpan.visitChildren skips spans with null text, so a styling-only root or nested span could supply shadows that were never detected. The parser now checks those spans recursively as well. Added root-span and nested-span pixel regressions: both failed before the change and pass now; the existing direct-shadow case still passes. Full suite: 638 passed; flutter analyze clean. Commit is signed and GitHub-verified.

marandaneto:
> @marandaneto all three addressed in [a8f45fa](https://github.com/PostHog/posthog-flutter/commit/a8f45fa4758d8a94b61a58216c9f0d73df56d184) — both blocking ones reproduced first, then fixed, with a regression test each that fails on the previous head.
> 
> The grapheme one was as you described: a range ending mid-cluster gets a box for the part before the cut and nothing after it, so the box list is non-empty and the old check sailed past it. Range ends are now checked against grapheme boundaries.
> 
> For the shadows I took the conservative option you offered rather than growing the rects, since that wouldn't help `except` — the escaping shadow there belongs to text outside the revealed window, so there's no bound to grow by.
> 
> 636 tests green, analyze clean.

nice, thanks, i pushed a [fix](https://github.com/PostHog/posthog-flutter/pull/581/changes/6b1ce5b612e9b101ac0b5728d979389b4375f6ae) and signed commits so its lgtm

marandaneto (approved):

lukas-roqqu:
Thank you for all your support @marandaneto 🙏 A lot of maintainers never reply to PRs or issues at all. You came back same day, every time. Really appreciate it.

PR #583, opened 2026-09-15
fix(surveys): honor choice option shuffling

lucasheriques:

Flutter receives `shuffleOptions` from the native SDKs but ignores it in the Dart renderer. Honor it for single- and multiple-choice surveys, moving toward survey feature parity across SDKs.

Shuffle regular choices once per question, keep Other last, and preserve order while selecting or typing. Configured choices and submitted values stay unchanged. This works with existing native SDK versions; no dependency on the native UI shuffle PRs.

Each question gets fresh widget state, so there is no live choice-update reconciliation. The `foundation.dart` import is required for `listEquals`; removing it fails compilation.

- Four enabled-shuffle widget cases failed before implementation; disabled cases passed. All 618 tests now pass, including 11 cases covering response values, question transitions, Other input, and empty/duplicate choices.
- Dart formatting, analysis, public API snapshot, and CodeScene pass.
- Android release APK and web example builds pass. Apple builds are not verified locally because Xcode/CoreSimulator components are mismatched.

## :pencil: Checklist

- [x] Reviewed the code and added behavior tests.
- [x] Public API unchanged.
- [x] Added a patch changeset with `pnpm changeset`.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted).

Codex used CLI tools and CodeScene. Native counterparts: PostHog/posthog-ios#811 and PostHog/posthog-android#772. No manual device testing.


posthog_flutter/lib/src/surveys/widgets/choice_question.dart
```diff
@@ -31,13 +33,56 @@ class ChoiceQuestionWidget extends StatefulWidget {
...
+      indices.map((index) => widget.choices[index]).toList(),
+    );
+    return [
+      ...(unchanged ? shuffled.reversed : shuffled),
+      if (openChoice != null) openChoice,
+    ];
+  }
+
+  @override
+  void didUpdateWidget(covariant ChoiceQuestionWidget oldWidget) {
```
turnipdabeets: [suggestion] I don't think this can run in the app. Each survey opens in its own sheet, each question gets a fresh widget (`choice_question_$_currentIndex`), and Flutter never updates a survey while it's showing. So the choices can't change under this widget. Could we drop this block and its test to keep things simpler?

posthog_flutter/lib/src/surveys/widgets/choice_question.dart
```diff
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
```
turnipdabeets, line 1: [nit] `material.dart` already gives us `listEquals`, so this import isn't needed.

turnipdabeets (approved):

